PR #221, opened 2016-03-21
Monitor_nD: fix unrestorable neutrons after PROP_Z0 call

### LINKED ISSUE #220: McStas-components: Monitor_nD removes events even if restore_neutron=1
For the flat disk and square geometry, `Monitor_nD` uses `PROP_Z0` which `ABSORB`'s events if they don't intersect with the xy-plane, making them unrecoverable even with the `restore_neutron` flag set to 1.

This leads to the following:
3d
`restore_neutron` is set for both monitors and the sample scatters in 4pi, but the left monitor kills all events in negative x direction.

Instrument file to reproduce:
```
DEFINE INSTRUMENT bug()

TRACE

COMPONENT source = Source_div(xwidth=0.04,yheight=0.04,focus_aw=0.2,focus_ah=0.2,
                              E0=2,dE=0.1)
AT (0,0,0) ABSOLUTE

// specify focus_aw and _ah so randec_target_rect_angular is used
COMPONENT sample = Incoherent(radius=0.015,thickness=0.0,focus_aw=360,focus_ah=180)
AT (0,0,2) ABSOLUTE
EXTEND
%{
   { // only scattered events
   if (!SCATTERED)
      ABSORB;
   }
%}

COMPONENT target = Arm()
AT (1,0,0) RELATIVE sample

COMPONENT mon_x = Monitor_nD(xwidth=1,yheight=1,options="x y",bins=50, restore_neutron=1)
AT (1,0,0) RELATIVE sample
ROTATED (0,90,0) RELATIVE sample

COMPONENT mon_y = Monitor_nD(xwidth=1,yheight=1,options="x y",bins=50, restore_neutron=1)
AT (0,1,0) RELATIVE sample
ROTATED (90,0,-90) RELATIVE sample

END
```

markusappel:
Fixes #220

### DIFF
--- a/mcstas-comps/monitors/Monitor_nD.comp
+++ b/mcstas-comps/monitors/Monitor_nD.comp
@@ -349,21 +349,37 @@ TRACE
        x,y,z, vx, vy, vz, &offdata );
     if (Vars.Flag_mantid) {
       if(intersect) {
-	Vars.OFF_polyidx=(offdata.intersects[offdata.nextintersect]).index;
+        Vars.OFF_polyidx=(offdata.intersects[offdata.nextintersect]).index;
       } else {
-	Vars.OFF_polyidx=-1;
+        Vars.OFF_polyidx=-1;
       }
     }
   }
-  else if (abs(Vars.Flag_Shape) == DEFS.SHAPE_SQUARE) /* square xy */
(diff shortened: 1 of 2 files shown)


willend:
@farhi this looks good to me, but could you have a look?

farhi:
I approve this proposal indeed. Thanks Peter and Markus !

PR #222, opened 2016-03-21
mccode-r.c: fix random vector distribution in randvec_target_rect_angular

### LINKED ISSUE #181: Runtime: randvec_target_rect_angular does not give uniform distribution over area
In `randvec_target_rect_angular` a random vector is drawn from a uniform distribution in spherical coordinates theta and phi:
```c
    /* Now choose point uniformly on quadrant within angle theta0/phi0 */
    theta = width*randpm1()/2.0;
    phi   = height*randpm1()/2.0;
```
... which will lead to increased probability at the poles. I *think* the correct way would be 
```c
    theta = width*randpm1()/2.0;
    phi   = asin(randpm1()*sin(height/2.0));
```
similar to `randvec_target_circle` (theta and phi seem to be inverse to their usual definition in `randvec_target_rect_angular`), with an additional check that `0 < height <= pi`.

markusappel:
fixes #181 

`randvec_target_rect_angular` is also defined in `mcxtrace/lib/share/mccode-r.c`.
Should this be changed as well?

### DIFF
--- a/common/lib/share/mccode-r.c
+++ b/common/lib/share/mccode-r.c
@@ -2742,8 +2742,8 @@ randvec_target_circle(double *xo, double *yo, double *zo, double *solid_angle,
 
 /*******************************************************************************
  * randvec_target_rect_angular: Choose random direction towards target at
- * (xi,yi,zi) with given ANGULAR dimension height x width. height=phi_x,
- * width=phi_y (radians)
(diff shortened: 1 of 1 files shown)


willend:
@markusappel this also looks good, plan to merge later today following a couple of tests I am currently performing.

PR #457, opened 2017-05-02
Update of Union example documentation section and absorber bug fix

mads-bertelsen:
Updated documentation section on all Union example instruments, but there is still work to be done on the descriptions and examples. This will however close the issue of duplicate instrument names in the example folder.

Found a bug in the special case of having no Union process before a Union_make_material component, which is allowed if one is making an absorber, it did however fail to compile. The Union_make_material component lacked a call to the Union_initialization.c script.

### DIFF
--- a/mcstas-comps/contrib/union/Union_make_material.comp
+++ b/mcstas-comps/contrib/union/Union_make_material.comp
@@ -6,6 +6,7 @@
 * %I
 * Written by: Mads Bertelsen
 * Date: 20.08.15
+* Version: $Revision: 0.1 $
 * Origin: Svanevej 19
 *
 * %D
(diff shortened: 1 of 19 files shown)


climbcat:
Hi Mads, super cool :-) 

A very few things: 
- The "Version: $Revision: 0.1 $" lines in the header have been deprecated.
- The stuff between "%INSTRUMENT_SITE" and "%D" or "%Description" is actually a short_desription which is displayed in the table generated by mcdoc (and elsewhere). In e.g. mcstas-comps/examples/Union_demonstration.instr and many of your other demos, you moved this short-description line into the description proper section. It would be helpful to retain these as a short_description.
- The format for params docs lines is: <parname>: [<unit>] <description>, where [] must be included even if the parameter does not have a unit par se. Many authors use [1] for pure numbers, or even a simple data type (such as int, string or char), but often it is left empty. 

We have just overhauled the header documentation, so currently all headers conform to this standard.

PR #548, opened 2017-09-29
Remove python plot error

ddatntu:
Error occurs when using the Python GUI to plot the results, complains files are missing.
The plotgraph.py and pqtgfrontend.py files need to be copied across when the distribution is created. Added files to tools/Python/mccodelib/CMakeLists.txt.


climbcat:
Hi ddatnu. Thank's for your submit and for reporting this!

PR #554, opened 2017-10-21
Update Cu2MnAl.laz

t-weber:
The structure factor for the Heusler given here is actually |F|, not |F|^2.


willend:
Sorry for not having noticed your pull request before, pulled to master now.

PR #605, opened 2018-04-13
Pull from master to Dresden-topdown

willend:


willend:
removing this attempt

PR #906, opened 2019-12-06
Fix output of the 2D detector in ASCII format

jkrueger1:
Only the first time the marker for the 2D I data format was set, whereas
the errors and N data where always separated by the markers.

Without these markers it is very difficult to separate data sets
generated by sending signal USR2 to the simulation program.


willend:
Hi @jkrueger1,

Good point, thanks for sending us the pull request which is now merged into master.

Peter

PR #953, opened 2020-01-28
changed default RNG to KISS

tweber-ill:


willend:
Tested and works, merging

Issue #911, opened 2020-01-13
GPUhack: Edit components for removal PGI clauses in cases like


willend:
<pre>
#ifndef USE_PGI
printf("Something...\n");
#endif
</pre>

(These things are now automatically removed by defines in the cogen part)

farhi:
PR https://github.com/McStasMcXtrace/McCode/pull/955 is about that

PR #955, opened 2020-01-28
PGI ifndef cleanup

tweber-ill:
Removed ifndefs around functions which are nullified in codegen for GPU. 
-> Issue #911


willend:
Hi @tweber-ill and @farhi,

It seems the #ifndef USE_PGI's in functions that are not TRACE itself are still needed, e.g. for ISIS_MERLIN and ViewModISIS when compiling for GPU I find that
<pre>
PGC-S-0155-Call to variadic function not supported - fprintf (ISIS_MERLIN.c: 6462)
PGC-S-0155-Accelerator region ignored; see -Minfo messages  (ISIS_MERLIN.c)
getPoint:
      0, Accelerator region ignored
   6461, Accelerator restriction: invalid loop
</pre>

So I guess we should give it another round tomorrow :-)

willend:
Now merged to mcstas-3.0

PR #957, opened 2020-01-28
Fix split in cpu runs

### LINKED ISSUE #954: GPUhack: SPLIT seems broken in CPU runs
@trnielsen noticed the CPU runs are coming in at a factor of 10 too low. 

Speculation: When implementing NOSPLIT, I may have done the /10 in SPLIT and removed the for-loop?

willend:
Fixes #954


willend:
Tested on multiple installations and works OK

PR #962, opened 2020-01-30
instrument.y: display meaningfull comp instance names

farhi:
Now display meaningfull error messages when a component parameter is not set correctly. A "(null)" was given as component instance name. Now more informative.

For instance we now have:
```
ERROR: Unassigned SETTING parameter l for component guide2=Elliptic_guide_gravity() at line Reflectometer.instr:302. Please set its value.
```
and it was before:
```
ERROR: Unassigned SETTING parameter l for component (null)=Elliptic_guide_gravity() at line Reflectometer.instr:302. Please set its value.
```


willend:
Nice, I will test and merge later today :-)

PR #965, opened 2020-01-31
Replaces any use of USE_PGI with USE_GPU and removes implicit NOSIGNA…

willend:
…L and RNG_ALG


willend:
I will merge myself after a succesful test, currently running

willend:
Hmm needs more work :-)

willend:
As we can now also target / build for -ta:multicore, I think a better name for this define would in fact be 
``` -DOPENACC```
Closing this PR and will do a fresh attempt on a new branch

PR #979, opened 2020-02-05
F funnel idxs

climbcat:


climbcat:
This changes the sort_absorb_last version to a) be more robust, and b) use indexes rather than pointers.

Issue #997, opened 2020-03-26
PowderN: concentric/nested geometries broken


markusappel:
Scattering events in hollow geometries with PowderN are not properly handled if the ray is already inside the void.

Example of a source in a hollow sphere:
```
DEFINE INSTRUMENT PowderN_nesting()

TRACE

COMPONENT Source = Source_simple(
  yheight  = 0.01,
  xwidth   = 0.01,
  dist = 0.5,
  focus_xw = 0.01,
  focus_yh = 0.01,
  lambda0 = 1.8,
  dlambda = 0)
AT (0,0,0) ABSOLUTE

COMPONENT analyzersupport = PowderN(
  reflections = "Al.laz",
  radius = 2.000,
  thickness = 0.001,
  p_interact = 1,
  p_inc = 0)
AT (0,0,0) ABSOLUTE
EXTEND
%{
  if (SCATTERED)
    printf( "Scattered at xyzt=[% .3f % .3f % .3f % .3e]  R=%.3f\n", x, y, z, t, sqrt(SQR(x)+SQR(y)+SQR(z)) );
%}

END
```

Result:
```
Scattered at xyzt=[-0.017  0.012  2.000  9.098e-04]  R=2.000
Scattered at xyzt=[ 0.010 -0.006  1.999  9.098e-04]  R=2.000
Scattered at xyzt=[-0.002  0.001  0.000  0.000e+00]  R=0.002
Scattered at xyzt=[-0.003  0.002  0.000  0.000e+00]  R=0.003
Scattered at xyzt=[-0.004  0.004  0.000  0.000e+00]  R=0.006
Scattered at xyzt=[ 0.009 -0.004  2.000  9.099e-04]  R=2.000
Scattered at xyzt=[-0.002  0.004  0.000  0.000e+00]  R=0.004
Scattered at xyzt=[ 0.017  0.005  1.999  9.097e-04]  R=1.999
Scattered at xyzt=[ 0.001  0.008  1.999  9.097e-04]  R=1.999
Scattered at xyzt=[-0.003  0.002  0.000  0.000e+00]  R=0.004
Scattered at xyzt=[-0.002 -0.028  1.999  9.098e-04]  R=1.999
Scattered at xyzt=[-0.003 -0.001  0.000  0.000e+00]  R=0.004
Scattered at xyzt=[ 0.000 -0.005  0.000  0.000e+00]  R=0.005
[...]

PR #998, opened 2020-03-26
PowderN.comp: fix scattering point in hollow shapes

markusappel:
Fixes the example given in issue #997 
Please triple-check if that doesn't break anything else, the PowderN component is a bit cryptic to me...


markusappel:
On second thought, this might not be the best solution yet;
If the ray starts inside the solid 'front' part of a hollow shape, sampling is probably biased because it's distributed 50/50 between front and back, but the path lengths are different.

ebknudsen:
I've pulled in your edit for now, but let's leave the bug report open if you're not quite convinced that it does give the right distribution.

Issue #1094, opened 2021-01-20
mcdisplay.py code does not create proper types of the monitors


Fahima-Islam:
mcdisplay.py code  for generating mantid instrument Definition File (IDF) produces  as many types as the number of components of the monitors. The type of the monitor in Mantid is defined with the number of the pixels in the horizontal and vertical direction and the width and height of the monitor. However the component of a specific type of the monitor is defined by it's location in the space. There can be multiple positions of the monitor for a specific type of the monitor. Hence there can be multiple components of the specific type of the monitor. The code should not produce the same number of the types of the monitors as the number of the components.

Fahima-Islam:
For a rectangular detector (monitor) the pixel coordinate system in mantid is defined as : left direction : x is negative, bottom direction: y is negative and back direction: z is negative. But the mcdisplay.py code produces opposite convention for these directions. 

see the example below:
```
<!-- Pixel for Detectors. Shape defined to be a (0.001m)^2 square in XY-plane with tickness 0.0001m -->
<type is="detector" name="pixel">
  <cuboid id="pixel-shape">
    <left-front-bottom-point y="-0.001" x="-0.001" z="0.0"/>
    <left-front-top-point y="0.001" x="-0.001" z="0.0"/>
    <left-back-bottom-point y="-0.001" x="-0.001" z="-0.0001"/>
    <right-front-bottom-point y="-0.001" x="0.001" z="0.0"/>
  </cuboid>
  <algebra val="pixel-shape"/>
</type>
```
from the following 
https://docs.mantidproject.org/nightly/concepts/InstrumentDefinitionFile.html

Fahima-Islam:
When defining the type of the monitor there is an attribute called ```idstepbyrow``` needed to define the amount to increase the ID number in the direction which is used to specify how the ID is filled (vertically or horizontally). That means, If the ID is filled horizontally (i.e. if the attribute ```idfillbyfirst``` is ```x```), then ```idstepbyrow``` should be the number of pixels in the horizontal direction of the monitor. Otherwise, ```idstepbyrow``` should be the number of pixels in the vertical direction of the monitor. 

But the mcdisplay.py code does not work accordingly. It specifies ```idfillbyfirst``` as ```x``` but ```idstepbyrow``` as the number of the pixels in the vertical direction. Which is not right.

Fahima-Islam:
[...]

PR #1105, opened 2021-01-31
mcdisplay.py code does not create proper Instrument Definition Files (IDF) that Mantid can load

Fahima-Islam:
Ref #1094 

This pull request corrected the mcdisplay.py code so that the detectors type will be unique (not repeated) in the IDF xml file.  It also use the angles are in "degree" rather than "radian".

This pull request also corrected `SNS_ARCS_Mantid.instr ` file so that the the event monitors are named as  “nD_Mantid_#” according to the [link ](https://github.com/McStasMcXtrace/McCode/wiki/McStas-and-Mantid) of how to write the McStas .instr file for generating Mantid IDF


Fahima-Islam:
@willend Can you please review this pull request? It should work now for McStas instrument file to generate the Mantid IDF.

PR #1195, opened 2022-01-21
Rock soleil

antoinepado:
# ROCK SOLEIL
## Generating the reflectivity files

Get the python script _reflec_xraydb.py_ located in data and run it in the command line: 

`python3 reflec_xraydb.py 3400 46300 0.0001 2 Pd 12.023`
`python3 reflec_xraydb.py 3400 46300 0.0001 2 Pt 21.45`
`python3 reflec_xraydb.py 3400 46300 0.0001 2 B4C 2.52`
`python3 reflec_xraydb.py 3400 46300 0.0001 2 Ir 22.56`

### Python script

To see the graph of the absorption coefficient as a function of the energy we scanned we can use the python script _absorption_fct_energy.py_ that can be found [here](https://pastebin.com/pkN3hM1d). Feed it the mccode.dat file. Simply run:
`python3 absorption_fct_energy.py`

### Directly in McXtrace (experimental/very possibly the wrong way of doing it)

We use a monitor_nD to plot the graph in McXtrace. 
We add the weights p to get the intensity of each monitor. The end value is not exactly the same as the intensity value of the Energy monitor. 
Indeed, the intensity value of the Energy monitor is the sum of all the elements of E_p. 
This is why the graph produced with the python script and the graph produced this way differ slightly.
We need to add all the values of E_p, how can this be done in the extend part? 

Is there a better/more correct way of creating a graph calculated from the intensity values of two monitors in McXtrace?


willend:
@antoinepado thanks, this looks good - will merge now.

Not sure if I understand your question fully, but one more slightly more direct way would be to define ```coeffabsorption``` as a  DECLARE variable (in McXtrace 3 this would be a USERVARS variable) to only depend on the independent particle, perform any calculation you need based on the particle weight and monitor it using a Monitor_nD with

```
COMPONENT blah = Monitor_nD(... signal=coeffabsorption ...)
```

@farhi @ebknudsen please do comment.

ebknudsen:
What you are doing right now @antoinepado is to compute the absorption coefficient for all x-rays on the fly, right? This is however just the outgoing intensity divided by the incoming, so you could simply extract this from the scan data-file if you had a energy-resolved monitor before and after. The integrated intensity of those two monitors are then represented by columns in the mccode.dat-file generated by the mxrun scanner. I think I'd do it that way. I may be misunderstanding you perhaps

PR #1214, opened 2022-02-23
SOLEIL_ROCK update

antoinepado:
Three changes to the SOLEIL_ROCK instrument have been made.

- The range of energy of the bending magnet source now  goes from 4 to 40 keV and we have `Ec = 0.665*B*Ee*Ee = 8.65 keV`

- A Monitor_nD has been added. It shows the y(height of the monitor) in function of the energy.

- Finally, we directly fetch the intensity, intensity error and number of events from monitors. 
With these values we calculate the absorption coefficient and its error.

### DIFF
--- a/mcxtrace-comps/examples/SOLEIL_ROCK.instr
+++ b/mcxtrace-comps/examples/SOLEIL_ROCK.instr
@@ -1,12 +1,11 @@
-
 /*******************************************************************************
 * Instrument: ROCK beam-line at SOLEIL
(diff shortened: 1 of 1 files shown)


farhi:
Perhaps there is a line swap around 103-104 in ROCK.instr ?
```
else if (Etohit<=34055){
        }
            cc = 0;
```
else the logic is strange. I would assume cc=0 is for E<34keV.

PR #1253, opened 2022-06-20
correct absorption cross-section

antoinepado:
Changed the absorption cross-section to:

`(mu[cm^-1]/rho[g*cm^-3])*(ma[g*mol^-1]/NA[mol^-1]) = sigma[cm^2]`

See [here](https://en.wikipedia.org/wiki/Absorption_cross_section).


willend:
Hi @antoinepado,

Thanks, will merge this in. Will you make the same correction for the mccode-3 branch?

Also, I guess that in principle this patch is also relevant on the neutron side?

PR #1260, opened 2022-07-21
Grating_reflect fix and czerny_turner monoch

antoinepado:
This is an initial fix of the Grating_reflect.comp component.
Will still need to take into account the case where the incident ray does not hit the grating in a perpendicular fashion.
Added is a test for the component.

Also added is a Czerny-turner monochromator based on [this](https://inis.iaea.org/collection/NCLCollectionStore/_Public/19/019/19019134.pdf).
Here is a [picture](https://imgur.com/a/jdMN11n) of a monitor that shows the Wavelength(Ang) as a function of x_screw(mm). To obtain it simply have the scan parameter be equal to 1 and scan from x_screw=20 to 103 mm in the Czerny-turner instrument.

### DIFF
--- a/mcxtrace-comps/examples/Test_grating_reflect.instr
+++ b/mcxtrace-comps/examples/Test_grating_reflect.instr
@@ -1,0 +1,119 @@
+/*******************************************************************************
+*         McXtrace instrument definition URL=http://www.mcxtrace.org
+*
+* Instrument: Test_grating_reflect
+*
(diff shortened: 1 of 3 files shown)


antoinepado:
Oh also, the grating used in the Czerny_turner monochromator is not blazed as of now. Will modify the instrument file to use a blazed grating at a later date.

Issue #1270, opened 2022-08-19
Floating point division causes Qt error


g5t:
For the latest `v3.1` release available for (at least) x86_64 linux and macOS, attempting to interactively configure an instrument via `mcgui` raises a `TypeError`.

## Steps to reproduce the error
1. Start `mcgui` and load an instrument `*.instr` file with user-configurable parameters.
2. Click on `Run...`

This results in an error message printed to `stderr` and a seemingly-hung GUI with the status message `Getting instrument params...`.

## Commandline output showing the error traceback
```shell
[~]$ mcgui 
loading system configuration
QSocketNotifier: Can only be used with threads started with QThread
Traceback (most recent call last):
  File "/usr/local/mcstas/3.1/bin/../tools/Python/mcgui/mcgui.py", line 596, in handleRunOrInterruptSim
    fixed_params, new_instr_params, inspect, mcdisplay, autoplotter = self.view.showStartSimDialog(
  File "/usr/local/mcstas/3.1/tools/Python/mcgui/viewclasses.py", line 157, in showStartSimDialog
    self.__ssd.createParamsWidgets(params)
  File "/usr/local/mcstas/3.1/tools/Python/mcgui/viewclasses.py", line 798, in createParamsWidgets
    self.ui.gridGrid.addWidget(lbl, y, x, 1, 1)
TypeError: arguments did not match any overloaded call:
  addWidget(self, QWidget): too many arguments
  addWidget(self, QWidget, int, int, alignment: Union[Qt.Alignment, Qt.AlignmentFlag] = Qt.Alignment()): argument 2 has unexpected type 'float'
  addWidget(self, QWidget, int, int, int, int, alignment: Union[Qt.Alignment, Qt.AlignmentFlag] = Qt.Alignment()): argument 2 has unexpected type 'float'

```

## Source of error
The layout-defining division
https://github.com/McStasMcXtrace/McCode/blob/0c15352d0d01015e96da2e93aa8c7e86a4e20833/tools/Python/mcgui/viewclasses.py#L792
results in a floating point value but an integer is required by `addWidget`.

## Proposed solution
Switch the floating point division for integer division, e.g., `y = i // (int(mccode_config.configuration["GUICOLS"])*2)`.

Implementing the proposed solution in the insitu `viewclasses.py` removes the `TypeError` and allows the `Start simulation` dialog box to appear as expectred.

willend:
Hi @g5t, thanks for both issue and proposed solution. :)

[...]

PR #1273, opened 2022-08-30
Internal accessors

g5t:
### Construct a component-name lookup table to identify `instrument` components
If two components rely upon each other, it may be useful to allow one to access properties of the other *including its position and orientation*. This functionality already exists via, e.g., the `POS_A_COMP_INDEX` macro and its friends, but it requires that the *entry-ordered* index for the target component be provided. Some components, like the `Source _simple`, use this machinery to determine the position and extent of *nearby* components, e.g., the **next** component, `INDEX_CURRENT_COMP+1`.
This machinery becomes cumbersome, however, if there are large number of components or the distance between components is large, since the possibility of counting errors scales with the number.

Instead a new mechanism, `_getcomp_index`, is provided to component authors which takes a component name and returns an integer appropriate to be used with `POS_A_COMP_INDEX` and the like.
Since component names must be unique, the returned number is therefore also unique.
A component that uses this functionality can therefore allow users the ease of providing a targeted-component *name* instead of needing to accurately count the possibly-changing number of components between the current and target components.

### Set particle parameters
Instruments can define `USERVARS` which are appended to the properties of the `_class_particle` struct. These properties can easily be retrieved in a component via the `particle_getvar` routine, but there is no straightforward way for a component to *set* the particle user properties. 

A new function, `particle_setvar`, is provided to remedy this issue. It requires that each user property be *assignable* from a `double` value passed-in, and therefore the thus-used `USERVARS` should **probably** *be* `double` valued.
In contrast to the variable retrieval routine, setting is limited to *only* the user-defined particle parameters.

## Fix floating point API errors
I have run-across two additional places in the Python code where integer values are expected by floating point values are provided. This PR fixes the two new issues in addition to (at least partly) addressing #1270.

### DIFF
--- a/common/lib/share/mccode-r.h.in
+++ b/common/lib/share/mccode-r.h.in
@@ -362,6 +362,8 @@ char  *mcsig_message;
 /* the body of this function depends on component instances, and is cogen'd */
 void* _getvar_parameters(char* compname);
 
+int _getcomp_index(char* compname);
+
 /* Note: The two-stage approach to COMP_GETPAR is NOT redundant; without it,
 * after #define C sample, COMP_GETPAR(C,x) would refer to component C, not to
 * component sample. Such are the joys of ANSI C.
--- a/mcstas/src/cogen.c.in
+++ b/mcstas/src/cogen.c.in
@@ -995,6 +995,27 @@ void cogen_getvarpars_fct(struct instr_def *instr)
   cout("");
 }
 
+/*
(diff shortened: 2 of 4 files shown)


willend:
Thanks for the patches! Independent work on the code generator in a first PR is actually quite an achievement! :-) 👍

farhi:
Congrats Greg !

PR #1277, opened 2022-09-15
[Refactor] - particle_setvar; + particle_*_void

g5t:
- The implementation of `particle_setvar` made the assumption that all user-provided `USERVAR` struct members would be `double` valued *or* could be transparently assigned from a `double`. This assumption was shown to be wrong by the example ILL_IN13.isntr which has a `Rotation` struct member, `R`. Generated `ILL_IN13.c` could not compile due to the inability to assign a double value to the `Rotation` member.
- One could imagine implementing separate `particle_setvar_[type]` functions with only the appropriate struct-members settable via each specialization, but this could lead to a proliferation of functions and requires that the user know the type of member. Instead, a new function `particle_setvar_void` takes a `void *` to the desired new-value and uses `memcpy` to make the assignment.
- This implementation does not *require* the user to know the type of each struct memeber, but providing the wrong data type will produce unexpected results.

- A matching function `particle_getvar_void` returns a `void *` to the requested data, relying on the user to cast the pointer to the appropriate type and dereference the cast-pointer to get its value. This allows accessing non-`double` particle struct member values.

- To go along with `particle_setvar_void`, a new function `particle_setvar_void_array` has been added which `memcpy`s more than one array element for array particle struct members. It does not know about (and therefore can not check the validity of) array memory limits.

### DIFF
--- a/mcstas/src/cogen.c.in
+++ b/mcstas/src/cogen.c.in
@@ -104,7 +104,7 @@ int parse_curlybrackets_vector(char* string, double* values) {
 
     // extract a value
     double val = strtod(s, &r);
-    if (!values==NULL) values[vidx] = val;
+    if (values!=NULL) values[vidx] = val;
     ++vidx;
 
     // iterate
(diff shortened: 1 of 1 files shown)


willend:
Tested to work as expected in GPU settings (i.e. ILL_IN13 compiles and runs as expected with this patch)

PR #1278, opened 2022-09-16
Support CMD() and ENV() in DEPENDENCY lines

tkittel:
As discussed with @willend, this PR injects a bit of dynamic features into the `DEPENDENCY` lines by introducing support for using environment variables or the output of commands via the keywords `CMD(..)` and `ENV(..)`. So for instance one might have dependency lines like:

```c
DEPENDENCY "-LENV(FOOLIBDIR) -lfoo"
DEPENDENCY "CMD(foo-config --cflags) CMD(foo-config --ldflags)"
DEPENDENCY "CMD(ENV(FOOPREFIX)/bin/foo-config --flags)"
```
Note that as the last example shows, one can use ENV(..) inside CMD(..), but apart from that there is no nesting allowed.

The immediate motivation for this feature is to make it easier to support usage of external projects such as NCrystal, where we would use the `ncrystal-config` script to query for the necessary flags. Thus, a user would merely need to activate a different NCrystal installation in order to make it the one used with the two McStas components which use NCrystal. That would do away with the need for my hacky workaround with the `ncrystal_preparemcstasdir` script which currently works for `NCrystal_sample.comp` but not the NCrystal process in McStas Union geometries. Hopefully it might find other uses as well down the line.

Concerning the details of my implementation, it is basically adding a new function, `evaluate_dependency_str` which parses a dependency string and as needed looks up environment variables (in os.environ) or runs commands via the subprocess module (capturing its stdout+stderr), and returns the final string after inserting the results. It is considered an error if an invoked command produces more than one non-empty line of text. The function handles both `str` and `bytes` types of arguments, and of course use the same type for the returned result. As a safety, the function immediately returns the argument unchanged if it finds neither `CMD(` or `ENV(` anywhere in the string. So hopefully it should be very unlikely that this introduces any unexpected problems on components not actually using these new features (which for now is all of the components).

As agreed, the feature is not added to the legacy-perl code. So in the future any McStas+NCrystal users who need to stick to the legacy-perl code will have to hack their DEPENDENCY lines manually.

To keep `tools/Python/mcrun/mccode.py` tidy, I added the new function in a new module `tools/Python/mccodelib/cflags.py`. If you instead wish me to add the function to `mccode.py` or in another existing module, just let me know and I will update the PR.

### DIFF
--- a/tools/Python/mccodelib/cflags.py
+++ b/tools/Python/mccodelib/cflags.py
@@ -1,0 +1,88 @@
+'''
+Utility function for handling flexible DEPENDENCY lines.
+'''
+import os
+import subprocess
+
+def evaluate_dependency_str( depstr, verbose=False ):
+    """Evaluates CMD() and ENV() parts of DEPENDENCY (a.k.a. "CFLAGS")
+    strings.
+
+    If neither "CMD(" or "ENV(" is found in the input string, the input
+    string will always be returned completely unchanged!
+
+    Input can either be bytes or str objects, and an object of the same type
+    will be returned.
+
+    The current implementation does not support usage of parentheses inside the
+    CMD(..) and ENV(..) blocks, nor nesting of these apart from the fact that
+    ENV(..) blocks can be used inside CMD(..) blocks.
+
+    """
+    #Make sure we can handle both raw bytes and str objects. We assume utf8 is
+    #sufficient for when we might have to convert output of environment
+    #variables or commands.:
+    is_bytes = lambda x : hasattr(x,'decode')
+    use_bytes = is_bytes( depstr )
+    if use_bytes:
+        to_target_fmt = lambda x : ( x if is_bytes(x) else x.encode() )
+    else:
+        to_target_fmt = lambda x : ( x if not is_bytes(x) else x.decode() )
+
+    s_ev = to_target_fmt('ENV(')
+    s_cmd = to_target_fmt('CMD(')
+
+    if not s_ev in depstr and not s_cmd in depstr:
+        return depstr #<--- early exit for most callers
+
+    as_str = lambda s : ( s.decode() if hasattr(s,'decode') else s )
+    s_endparan = to_target_fmt(')')
+
+    def evalmarker( s, startmarker, evalfct ):
+        if not startmarker in s:
+            return s
+        before, ca = s.split(startmarker,1)
+        if not s_endparan in ca:
+            raise ValueError( 'Missing closing parenthesis in dependency '
+                              +f'string after opening "{as_str(startmarker)}"' )
+        content, after = ca.split(s_endparan,1)
+        if startmarker in content:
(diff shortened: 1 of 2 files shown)


willend:
@tkittel thanks for this, much appreciated!

One thing that has just come to my mind is a minor downside: 
What one writes in a CMD should of course be "platform independent"... 

(Which means that for some Unix/POSIX things the component developer may need to do some .bat scripting etc. on Windows... Or simply declare that the component does not work there... :-) )

tkittel:
Thanks @willend!

Concerning the platform dependency issue you bring up: An easy fix with the way the code is laid out would be to add platform-specific variants of `CMD()`: `CMDWIN()`, `CMDOSX()`, `CMDLINUX()`, `CMDUNIX()` (unix meaning "not windows"). These would be like `CMD(..)` when the platform matches, but will simply be replaced with an empty string when it does not.

So one could imagine:
```c
DEPENDENCY "CMDWIN(foobar /some /win /command)CMDUNIX(foobar --some --unix --command)"
```
I guess we would define the windows WSL as being LINUX and not WIN for the sake of this.

Do you want me to prepare another PR adding support for  these CMDxxx()?

willend:
@tkittel thank you, will think about that one. :) 

Honestly I think the world would be a simpler place if everyone just ran *nix... ;-)

Since you are for now the "only customer" for these new DEPENDENCY features, I think we should simply leave as is. (for now at least) :-)

tkittel:
Ok, probably good to think about it a bit longer. I just realised that the platform-specificity might in any case not be best handled via specialisations of `CMD()` since it might be a non command that is different (e.g. needs `-lFOO` on windows and `-lfoo` on unix). So we could add `ONLYWIN`, `ONLYUNIX`, etc. which would be evaluated _before_ `CMD(..)` and `ENV(..)`. On the indicated platform they would be replaced with their arguments, and on other platforms by empty strings. So one might do:
```c
DEPENDENCY "ONLYWIN(-lFOO)ONLYUNIX(-lfoo)"
```
And the example from my last comment could then be achieved with either of:
```c
DEPENDENCY "ONLYWIN(CMD(foobar /some /win /command))ONLYUNIX(CMD(foobar --some --unix --command))"
DEPENDENCY "CMD(foobar ONLYWIN(/some /win /command)ONLYUNIX(--some --unix --command))"
```
In any case, I am hoping that invoking the NCrystal configure script on windows like `ncrystal-config --show libdir` would work just as well as it does on unix. Until I somehow find out otherwise, this issue is indeed a bit academic. Also because I am not 100% sure whether or not NCrystal can even currently be built on windows ;-)

willend:
What I know is that at least I don't seem to build NCrystal when packaging McStas for Windows (using my cross-compilation approach), so those that have it (if any) must have found their own way. :-D

tkittel:
Yeah, you are right, it is definitely a non-issue for now :-)

Issue #1299, opened 2022-11-07
Building using CMake requires external libraries, e.g., MCPL, to be installed beforehand


g5t:
These libraries need to be installed into a subdirectory `3rdparty` before running CMake.
A script is provided to install the libraries, but we can use built-in CMake functionality `FindPackage` and `FetchPackage` to preferentially use a pre-existing copy of each library or pull a copy from, e.g., a GitHub repository.

Change the interface to third-party code to use CMake functionality

PR #1303, opened 2022-11-07
[Refactor] CMake should pull-in MCPL and NCrystal

g5t:
Both MCPL and NCrystal are built along side McStas & McXtrace. These
should only be pulled from a remote repository and built if they are not
present locally.
This commit adds CMake modules to find and fetch if not present any
general CMake project, plus to do so for the MCPL and NCrystal projects.

The CMakeList files are updated to provide new option flags to build
either. The old behvior of requiring externally-populated 'thirdparty'
directories is maintained by default.


g5t:
This PR would close #1299

willend:
Yup, seems to work, looks good. Tested with NCrystal_example.instr and Test_MCPL_input.instr (the latter locally modified, should use the new mcpl config stuff in DEPENDENCY)

Issue #1249, opened 2022-06-08
Single_crystal suppresses parameter-input by reflection-file content.


willend:
Observations from @antoinepado and @mads-bertelsen indicate that both the McStas and McXtrace Single_crystal have an issue with correctly defining (e.g.) lattice-orientation via the input-parameters.

In other words, it seems as if lattice-oriented content of the reflection-file likely overwrites user-provided inputs.

A good discussion and a major overhaul of the code / logic is envisioned.

willend:
I believe the interface logic was since corrected / debugging output printed from the comp.

PR #1305, opened 2022-11-08
[Add] UNSET macro to create special NaN

g5t:
- Currently there is no special value used for component parameters which are optional or mutually exclusive. Component authors may use an unphysical value in some cases, but there are examples where no suitable value exists.
- A new macro, `#define UNSET nan("0x...")`, allows components to use the special NaN value to keep track of whether a user has provided any input for the specific parameter(s). Helper functions have been added in the form of `is_unset`, `is_set`, and `is_valid`, to check whether a specific double value is the special NaN. The standard Slit.comp component has been modified to use the new NaN value, and now has logic at initialization to verify that a consistent set of parameter(s) have been provided.
- The specific NaN chosen for UNSET should not be created as the result of bad calculations elsewhere, but if it is that may lead to confusion for component authors and users. This possiblity is left for further consideration in the future.

### DIFF
--- a/common/lib/share/mccode-r.c
+++ b/common/lib/share/mccode-r.c
@@ -117,6 +117,21 @@ size_t str_len(const char *s)
 
 #endif
 
+/* SECTION: Predefine (component) parameters ================================= */
+
+int nans_match(double a, double b){
+  return (*(uint64_t*)&a == *(uint64_t*)&b);
+}
+int is_unset(double x){
+  return nans_match(x, UNSET);
+}
+int is_set(double x){
+  return !nans_match(x, UNSET);
+}
+int is_valid(double x){
+  return !isnan(x)||is_unset(x);
+}
+
 
 /* SECTION: Dynamic Arrays ================================================== */
 IArray1d create_iarr1d(int n){
--- a/common/lib/share/mccode-r.c
+++ b/common/lib/share/mccode-r.c
@@ -131,6 +131,38 @@ int is_set(double x){
 int is_valid(double x){
   return !isnan(x)||is_unset(x);
 }
+int all_unset(int n, ...){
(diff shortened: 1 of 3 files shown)


g5t:
A simple instrument to exercise some of the new parameter logic in the updated Slit component follows:
```
DEFINE INSTRUMENT test_slit()
DECLARE
%{
%}
INITIALIZE
%{
%}

TRACE

COMPONENT Origin = Progress_bar() AT (0,0,0) ABSOLUTE

COMPONENT radial_slit = Slit(radius=3.0) AT (0, 0, 1) RELATIVE PREVIOUS
COMPONENT symmetric_slit = Slit(xwidth=1.0, yheight=2.0) AT (0, 0, 1) RELATIVE PREVIOUS
COMPONENT offset_slit = Slit(xmin=1, xmax=3.0, ymin=-0.1, ymax=5.0) AT (0, 0, 1) RELATIVE PREVIOUS
COMPONENT closed_width_slit = Slit(xwidth=0.0, ymin=3.0, ymax=5.0) AT (0, 0, 1) RELATIVE PREVIOUS
COMPONENT closed_y_slit = Slit(xwidth=0.0, ymin=5.0, ymax=5.0) AT (0, 0, 1) RELATIVE PREVIOUS

FINALLY
%{
%}
END
```

g5t:
This commit will enable the transformations required by #1249

mads-bertelsen:
Looks good, with the new variadic functions it is much easier to read the input sanitation.

The Slit code would probably be easier to read if it handled everything related to x, then everything related to y, instead of having them intertwined.

It is a slight concern this can only work for doubles, but that is the vast majority of cases where this is relevant.

willend:
Tried it out, looks good, pulling it.

PR #1307, opened 2022-11-08
Work on CMake for MCPL/NCrystal + start to clean EXE suffix + toolchain files

tkittel:
To bring in sync before concurrent development with @willend takes us too far apart ;-)

I still need to work on getting everything to work without a toolchain file, and I would be surprised if there would not be some changes needed after this in the various build scripts, but probably that is best looked at tomorrow in the same room.


willend:
For both of McStas and McXtrace I have now built and tested for `cmake ..` and `./buildscripts/` type builds and most things look OK.

Pulling in as is. A little work is needed on my own side wrt. the MCPL_input instruments + localisation of the voutput.mcpl input file which seems to have become slightly different in the two codes.

PR #1308, opened 2022-11-09
Union clean up

mads-bertelsen:
Clean up of header on all Union components so they are read by mcdoc. Only PowderN and Single_crystal processes have issues at this point, any comments on how that could be rectified are welcome.

Clean up of Union_master in terms of outdated comments and unused features removed, namely writing absorption information to file which can now be done by the absorption logger components.


willend:
Looks good, have run a couple of tests on Union instruments and looked at most mcdoc comp pages

PR #1311, opened 2022-11-09
Make CMake cfg work without toolchain file

tkittel:
If variables are not set (like with a toolchain file), the CMake code will now try to come up with sensible defaults. This should hopefully make a bare-bones `cmake -DCMAKE_INSTALL_PREFIX=...` installation work out of the box.


willend:
Looks nice

PR #1312, opened 2022-11-10
Move PlatformDefaults inclusion to MCUtils.cmake

tkittel:
Move PlatformDefaults inclusion to MCUtils.cmake where it should have been. Also adds a BUILDS_COMPS flag (which is ON by default) to the top-level CMakeLists.txt


willend:
Great, thanks!

PR #1313, opened 2022-11-10
Traceless optimization

mads-bertelsen:
Optimization of the code generator that removes coordinate transformations from components if all of the following is true:

- Has no trace code
- Has no extend code
- Does not jump anywhere
- Is not jumped to

The component is skipped in the transformation chain too, so the transformation is still correct when the rays goes over a sequence of skipped components.

Introduced list access function in List, the review should check the index is handled correctly. It turned out components where 1 indexed in the code generator, but I don't think that is general for all lists, and so I made the index change in the code gen optimization.

### DIFF
--- a/mcstas/src/cogen.c.in
+++ b/mcstas/src/cogen.c.in
@@ -2813,48 +2813,59 @@ int get_codeblock_vars(struct code_block *code, List vars, List types,
   return(ans);
 } // get_codeblock_vars
 
-
+/*******************************************************************************
+* detect_skipable_transforms: Finds components where coordinate transform can be skipped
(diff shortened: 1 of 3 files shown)


mads-bertelsen:
Benchmark with useless arms:
image

Benchmark with big Union system: 
image

It is expected the Union system should run slower with larger systems as more intersection checks are needed. The 6 LOKI detector rows instrument has 302 components.

willend:
Looking at http://new-nightly.mcstas.org/2022-11-10/overview.html and the actual code edits this is very light and seems to break nothing. - So pulling in to mccode-3

PR #1315, opened 2022-11-10
[Fix] mcrun binary execution logic

### LINKED ISSUE #1314: `mcrun` with no points and trace should run the binary
In order to convert instrument geometry to external static formats we must export the geometry from the compiled instrument binary.  There are Python functions to parse the geometry information, so they should be re-used. They call `mcrun` with the  `-t/--trace` option to force output of both particle path *and* instrument geometry information.

External geometry formats like STEP can not hold particle path information, so the binary *could* be called with no particles requested in order to obtain all required information. This does not work, however, since `mcrun` does not call the binary if `-n/--ncounts` is zero, independent of any other options.

The logic controlling whether the binary is called should also consider the presence of the `--trace` flag. If it is present the binary should be called independent of the number of requested particles.

This has the advantage of solving the `mcdisplay-*` infinite hangs caused by there being no binary output to parse, and the parser waiting indefinitely for sub-process output that will never occur.

g5t:
Closed #1314 by adding a check for trace being turned on to allow running the binary with no requested particles.

### DIFF
--- a/tools/Python/mcrun/mcrun.py
+++ b/tools/Python/mcrun/mcrun.py
@@ -427,7 +427,7 @@ def main():
         scanner.run()
     else:
         # Only run a simulation if we have a nonzero ncount
-        if not options.ncount == 0.0:
+        if options.ncount != 0.0 or options.trace:
             mcstas.run()
 
     if isdir(options.dir):


willend:
:) thanks 👍

PR #1316, opened 2022-11-11
1309 output instrument geometry to step

g5t:
Tooling to ensure [CadQuery](https://cadquery.readthedocs.io) is available does not exist in this PR. But it maybe be better to add runtime output if `mcdisplay-cadquery` is called and the underpinning packages are not present.

Also up for debate: what to call the new tool. Options include,
- [ ] `mcdisplay-cadquery`
- [ ] `mcdisplay-step`
- [ ] `mcdisplay-cad-instrument`


g5t:
The branch from which this PR originanted needed to be rebased on the latest mccode-3 *before* the PR was created.
A new PR will be created from an appropriately based branch.

PR #1317, opened 2022-11-11
McDisplay improvements

g5t:
- Common parts of the `mcdisplay-` tooling have been coalesced into `mccodelib/mcdisplayutils.py` .
- `mcdisplay-webgl` has been fixed to *copy* files rather than reading and rewriting them in text mode
- A new `mcdisplay-` has been added that _can_ be used to output STEP files from instrument geometries.

Tooling to ensure [CadQuery](https://cadquery.readthedocs.io/) is available does not exist in this PR. But it maybe be better to add runtime output if mcdisplay-cadquery is called and the underpinning packages are not present.

Also up for debate: what to call the new tool. Options include,
1.  mcdisplay-cadquery
2. mcdisplay-step
3. mcdisplay-cad-instrument


g5t:
The last commit also adds pre-checking for the presence of the Python module `cadquery` and a command line option to set the output file type, validated to be [one of those supported by `cadquery.Assembly`](https://github.com/CadQuery/cadquery/blob/03c3266b8d61d487349a15f73d3b3cb13bce625f/cadquery/assembly.py#L456)

willend:
Tried and tested, now pulling it in. It seems cadquery is not immediately/easily available on all platforms, so I may decide not resolving that dependency "everywhere" for the next release.

willend:
Yikes, on Debian / Ubuntu the advice is to use snap... https://snapcraft.io/install/cadquery-editor/debian  :-D 

And on macOS/Silicon FreeCAD is a dependency and is so far not available in native form...

Until this clears up somehow (or McStas becomes available in conda) I will leave it to the users to deal with these dependencies.

PR #1320, opened 2022-11-11
[Fix] configuration error for mcdisplay-matplotlib

g5t:


willend:
This should've been done ages ago... Seriously error-prone...

Issue #1381, opened 2022-12-29
Problem in monitor_nD banana geometry


mads-bertelsen:
When Monitor_nD is used with the banana geometry, intersect is still done with a cylinder that contains the top and bottom. In the following code the intersect variable contains the return value from cylinder_intersect, this return variable will be 1 if only intersections with the curved part are found and larger if intersections with the top or bottom is found.

Intersections with the top and bottom needs to be removed and the following code attempts to remove them, but does not seem to work in all cases.

https://github.com/McStasMcXtrace/McCode/blob/62c86860a926668edf12e61fa4760e660152fbdd/mcstas-comps/monitors/Monitor_nD.comp#L479-L511

Cases that can create problems:
1. The intersection routine has some error which is taken into account on line 505, but not in the checks on line 491 and 492.
2. Ray starts outside and intersects with both top and bottom plate. Both the checks on line 491 and 492 are true, and the t0 
3. Error on intersection calculation is larger than FLT_EPSILON and in the direction of the inside of the cylinder, then the point is not removed even if it is not located at the edge.

As a temporary measure I have used a larger value than FLT_EPSILON and included this in the checks on line 491 and 492, then  the problematic data from the top and bottom disappears. This does remove a tiny fraction of the data from the desired part of the detector.

A proper fix could be a dedicated banana_intersect function that simply is a copy of the cylinder_intersect function without the flat parts, removing the need for the flaky filter code afterwards.

The Monitor_nD code is the same for 2.X and 3.X branches and affect both. 

#1318 could be used to test. Relevant test cases are beam hitting top and bottom and/or source being outside the cylinder.

mads-bertelsen:
The issue is further complicated by the use of gravity as PROP_DT will propagate with gravitation, where the temporary point check previously used linear propagation.

mads-bertelsen:
I have chosen another temporary fix as some edge cases still had too large numerical uncertainty and required relatively large floating point error buffers to fix. Instead I am using the return value from cylinder_intersect to filter the events directly, as it contains information on which face the ray intersected. The code is pasted below, can make a pull request on it if this is the desired direction, though there are a lot of cases to consider.

```
if (intersect)
  {
    if ((abs(Vars.Flag_Shape) == DEFS.SHAPE_SPHERE) || (abs(Vars.Flag_Shape) == DEFS.SHAPE_CYLIND)
[...]

PR #1384, opened 2023-01-02
Change of logic in Monitor_nD for Banana shape

mads-bertelsen:
Before the component checked if the intersection position was on the top/bottom of the cylinder, yet this requires comparisons of doubles and the result of the intersection calculation has some error, making it flaky. Some rays leaving through the top/bottom of the cylinder were included in the results.

Updated to use the return value from the intersection function to identify whether or not to keep events.

Discussed further in issue: #1381

### DIFF
--- a/mcstas-comps/monitors/Monitor_nD.comp
+++ b/mcstas-comps/monitors/Monitor_nD.comp
@@ -483,23 +483,53 @@ TRACE
      || (geometry && strlen(geometry) && strcmp(geometry,"0") && strcmp(geometry, "NULL")) )
     {
       /* check if we have to remove the top/bottom with BANANA shape */
-      if ((abs(Vars.Flag_Shape) == DEFS.SHAPE_BANANA) && (intersect != 1)) {
-        double y0,y1;
-        /* propagate to intersection point as temporary variable to check top/bottom */
-        y0 = y+t0*vy;
-        y1 = y+t1*vy;
-        if (fabs(y0) >= Vars.Cylinder_Height/2) t0 = t1;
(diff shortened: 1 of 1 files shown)


mcstas-comps/monitors/Monitor_nD.comp
```diff
@@ -483,23 +483,53 @@ TRACE
...
+            /* t0 is now time of incoming intersection with the detection area */
+            if ((Vars.Flag_Shape < 0) && (t1 > 0))
+              PROP_DT(t1); /* t1 outgoing beam */
+            else
+              PROP_DT(t0); /* t0 incoming beam */
+        }
+      
       /* Final test if we are on lid / bottom of banana/sphere */
       if (abs(Vars.Flag_Shape) == DEFS.SHAPE_BANANA || abs(Vars.Flag_Shape) == DEFS.SHAPE_SPHERE) {
         if (Vars.Cylinder_Height && fabs(y) >= Vars.Cylinder_Height/2 - FLT_EPSILON) {
```
mads-bertelsen, line 535: We could remove the FLT_EPSILON on this check for BANANA now, as it is effectively just checking if the path propagated with gravity sent the neutron outside of the detector area.
I don't understand the need for lid/bottom of the sphere mentioned in the comment, so have kept this code.

willend:
Thanks for the PR @mads-bertelsen. I have been running a quick test, visible as the column McStas_CPU_MPICC_KISS_5e6   in https://new-nightly.mcstas.org/2023-01-05/overview.html

Will now merge the patches into both master and mccode-3

PR #1385, opened 2023-01-02
Change of logic in Monitor_nD for Banana shape

mads-bertelsen:
Before the component checked if the intersection position was on the top/bottom of the cylinder, yet this requires comparisons of doubles and the result of the intersection calculation has some error, making it flaky. Some rays leaving through the top/bottom of the cylinder were included in the results.

Updated to use the return value from the intersection function to identify whether or not to keep events.

Fix to issue #1381


willend:
Merging into master

PR #1389, opened 2023-01-05
Update plotfuncs.py for numpy compatibility

jorobledo:
change np.float to float since np.float is deprecated.


willend:
Sounds and looks very reasonable! :) Merged in. (I will cherry-pick to mccode-3 also since the tools are ~ identical between master 2.x and mccode-3 3.x branches.)

PR #1405, opened 2023-01-25
Refactor python mctest to handle `Example` scans

g5t:
This is a first draft to implement instrument test cases which include parameter scan(s).

The test specification grammar is extended to allow for vector intensity (as well as error; which remains unused), via, e.g.,
`%Example: -N3 angle=1,3 Detector: tube_I=[100, 39, 10] tube_ERR=[10, 5, 3]`

The test-report generator is refactored to handle multi-valued comparisons -- at present this is done by finding the sum of both the simulated and expected intensities, but this could be extended to a more-useful comparison.

## Possible issues
If one wants to specify a _large_ scan, the test case may become very long, which could be problematic for non-line-wrapping editors. To help avoid any such issues, test cases can now be split over multiple lines using the stitching (or line continuation) operator `\` as the last character on a line.
When searching for test cases `\\n` is first replaced by a single space such that the logical line can be split over multiple lines in a file.
This functionality *only* works within C-style (`/* ... */` block) comments since the line-continuation functionality is not implemented in McStas for C++-style (`// ...` inline) comments.


g5t:
An important note if you'd like to test this: the scan-reading functionality within this relies on #1404 -- otherwise it only ever reads the first two points of any scan.

willend:
Tested functional, a nice little addition. Merging in.

willend:
Was reverted for now, sorry :)

PR #1414, opened 2023-02-24
Add check for Union_init component name to Union components

g5t:
Use of `Union` in `McStas` 3.2 requires a `Union_init` component to create shared memory data structures.
Most other `Union` components then access the shared memory via the `COMP_GETPAR3` macro which takes the component instance type, the component instance name, and the name of the component parameter to retrieve.

The component instance name must be provided, and to make use of `Union` easier a default name of `"init"` is assumed in every component making use of the `COMP_GETPAR3` macro.
If the `Union_init` component is created with a name other than `"init"` _and_ the other components are not configured with the different name, the compiled instrument **will likely** crash at runtime due to its access of out-of-process memory.

By adding a check for the component name existing, via the `_getcomp_index` function, a user can be alerted to the name-mismatch *before* the running program crashes allowing for easier pinpointing of the problem.
The main potential drawback to this implementation is if there is a component with the right name but the wrong type, in which case the running program will likely still crash without any useful error message.

This proposed change has not been tested yet, and should not be merged at this time.

### DIFF
--- a/mcstas-comps/union/IncoherentPhonon_process.comp
+++ b/mcstas-comps/union/IncoherentPhonon_process.comp
@@ -4669,8 +4669,15 @@ INITIALIZE
   sprintf(global_process_element.name,NAME_CURRENT_COMP);
   global_process_element.component_index = INDEX_CURRENT_COMP;
   global_process_element.p_scattering_process = &This_process;
-  
-  struct pointer_to_global_process_list *global_process_list = COMP_GETPAR3(Union_init, init, global_process_list);
+
+  if (_getcomp_index(init) < 0) {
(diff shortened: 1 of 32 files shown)


willend:
Thanks @g5t, will see if I can manage to give this a spin within the next 1-2 weeks, I guess @mads-bertelsen might also do some testing.

For the long term I am considering how global memory structures could potentially be put in place in the instrument / root level, which should then remove the need for Union_init and Union_stop.

g5t:
My other PR #1419 would allow an extra-safety check by *hiding* all non-init Union components from `mcstas` if the `Union_init` components has not been encountered in the instrument.

willend:
Merged after successful test (2e7 column in https://new-nightly.mcstas.org/2023-03-25/2023-03-25_output.html)

PR #1427, opened 2023-04-14
Update mcrun.py

farhi:
Add --optimize option


farhi:
Can be tested with e.g.:
```
mxrun --optimize -n 1e5 Test_SX TTH=13,14
```
with both McStas and McXtrace. The 2 files mcrun.py optimisation.py in `tools/Python/mcrun` should be upgraded.

willend:
Tested functional, merging in. Merge to mccode-3 pending.

PR #1428, opened 2023-04-14
Update optimisation.py

farhi:
Add Optimizer class for parameter optimization. Goes together with farhi-patch-1 PR #1427


willend:
Tested functional, merging in. Merge to mccode-3 pending.

PR #1429, opened 2023-04-18
Mccode 3 mcpygen - alternative, minimal code-generator creating .py equivalent of .instr

willend:
The prototype added here adds a small code-generator (`mcpygen`) which may generate a python (aka McStasScript) model from an existing instrument file.

Usage:
1. `mcpygen ILL_H5.instr` generates a python script (`ILL_H5_generated.py`)
2. `python3 ILL_H5_generated.py` generates `ILL_H5_generated.instr` and runs McStas

Current caveats are at least these two:
1. Instrument files using an input parameter `lambda` should instead use `Lambda` (corrected in attached version of `ILL_H5.instr`)
2. `CPU COMPONENT` grammar is not in place.

- But most of the general McStas grammar is functional

[ILL_H5.instr.txt](https://github.com/McStasMcXtrace/McCode/files/11265538/ILL_H5.instr.txt)


willend:
@mads-bertelsen I am taking the liberty of simply merging this in as there should be no implications on existing "production" code.

willend:
Oops. My bad, not fully integrated in build system so has broken last nights test… 😊 

Note to self: Last minute “I’ll just merge that” changes to the cogen infrastructure is a bAd idea… Will tidy during Wednesday.

willend:
Final version of the tool is called `mcstas-pygen` (and `mcxtrace-pygen` for McXtrace). 

After some fiddling today it seems the prototype is properly in place on `mccode-3`.

PR #1432, opened 2023-05-02
GPU compatible models from sasmodels update

jorobledo:
Updated existing models and added new models that are available at [sasmodels.models](https://github.com/SasView/sasmodels/tree/master/sasmodels/models). OpenACC compatibility was included, therefore all models may be deployed to GPU.

There are a few models that need to have an array as input parameter, like `sas_spherical_sld.c`, `sas_core_multi_shell.c`, `sas_onion.c`, that are still not available since the `sasview_proxy.c` file is not compatible with this type of input and they need a different logical implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@
 *~
 \#*\#
 .\#*
+
+3rdparty/
+Mambaforge-Linux-x86_64.sh
+dist/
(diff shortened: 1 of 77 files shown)


willend:
I've added a few models with %Example added in  https://github.com/McStasMcXtrace/McCode/commit/8ba6e7cebf963b6f719202fec13dd17c020409f2

(Those without % in the Example line only have issues for now.)

But clearly functional enough to merge in.

@jorobledo do you potentially have a script / instr file that I could inherit from for a more complete model coverage?

jorobledo:
Yes, I will push the model generator soon and I can also see to add at least one test example for all of the models if you find that useful. Also I thought on solving the "array as input parameter" problem in the next pull request.

PR #1434, opened 2023-05-05
Fixed DivE_monitor.comp

jorobledo:
It seems it used to use wavelength as input parameters, and some names were left unchanged, making the component not working. Changed the names accordingly and tested that it is working again.


willend:
Thanks @jorobledo, merging in!

@farhi maybe we need another bump of the installation(s) for the SOLEIL school? (I will immediately check if the same issue is on master.)

willend:
Similar edits merged to master also @farhi, @jorobledo

PR #1438, opened 2023-05-08
[Add] optional instrument parameter units

g5t:
# Current status
## Split definition of a parameter and its units
Instrument parameters must be specified in the instrument file following the name of the instrument, enclosed in parentheses.
Often a comment block precedes the `DEFINE` statement, which can be used to indicate the units of the parameters:

```
/********************************************************************
*
* Instrument: InstrumentName
*
* %Parameters
* parameter1: [m]    a length used in the instrument
* parameter2: [1]    a flag indicating some choice in the instrument setup
* parameter3: [rad]  the local value of pi
*
************************************************************************/
DEFINE INSTRUMENT InstrumentName(
            parameter1,          // required at run-time, double valued
    type    parameter2,          // required at run-time, type one of int, double, or string
    double parameter3 = 3.14159  // optional at run-time
)
...
```
The comment blocks and instrument definitions can be parsed to produce useful user documentation with `mcdoc`.

## Parameter units not available from binary
While there is an easy way to extract parameter types and default values (via `instrument_binary --info`), there is no equivalent way to extract the parameter units from the compiled instrument.

While not strictly necessary to use a compiled instrument, it would be nice to provide a means by which to convey the _intended_ units of a parameter to a user.

# Changes in this pull request
## Additional McCode grammar
The McCode grammar is extended to support optional string valued unit names for all instrument parameter types. 
Under the new syntax the parameters of the instrument above can optionally be made explicit

```
DEFINE INSTRUMENT InstrumentName(
            parameter1 ("m"),            // required at run-time, double valued
    int    parameter2,                  // required at run-time, here int replaces type
    double parameter3 ("rad") = 3.14159  // optional at run-time
    
)
...
```
The unit can be any valid `TOK_STRING` and therefore must be enclosed in double quotes.

## New generated runtime option
The code generator is extended to produce functionality to print complete parameter information from the compiled instrument.
If the above instrument was compiled to a binary, `instrument.out`, then its parameters would be accessible via the flag `-p` or `--params`:

```cmd
$ ./instrument.out -p
  double parameter1 ("m") = NULL
  int parameter2 = NULL
  double parameter3 ("rad") = 3.14159
```

# Possible issues
## Negative effect on `mcformat`?
The code generator appends the provided unit strings to the `mcinputtable_struct` and associated `mcinputtable[]` list.

It appears that `mcformat` also makes use of `mcinputtable[]`, and reads information from files to populate the parameter fields. Since what `mcformat` (presumably) reads is unmodified (the output of `mcinfo()` in the binary) the parsing code will probably still work. Also, since `mcformat` directly assigns members of the struct, it should continue to work with the struct redefined to include an extra `int` and `char *`.
Still,	these modifications might have an effect on the runtime use of `mcformat`, which should be checked.

## Less-than-ideal syntax
Specifying the units of a parameter as a literal string is good, as it can support all types of units likely to appear in an instrument, e.g., `1/m`, `\AA`, `minutes of arc`, etc.

Placing those units, however, within parentheses is not as good as, e.g., separating them from the parameter name by a forward slash. Compare `double parameter3 ("rad") = 3.14159` to `double parameter3/"rad" = 3.14159`, the latter of which makes it clear that the number (`3.14159`) should be interpreted as a number of `"rad"` to know the value of `parameter3`.

For reasons that remain unclear to me, I could not expose `"/"` as a singular token in `instrument.l`. Removing it from the `TOK_CTOK` list and adding a new rule to the 'operators' return and to `topatexp` `instrument.y` did not allow for the singular forward slash to be used in other rules, e.g., `TOK_ID '(' TOK_STRING ')' '=' exp` works but `TOK_ID '/' TOK_STRING '=' exp` does not.
Any help here would be greatly appreciated.

### DIFF
--- a/common/lib/share/mccode-r.c
+++ b/common/lib/share/mccode-r.c
@@ -1114,8 +1114,8 @@ static void mcinfo_out(char *pre, FILE *f)
     snprintf(ThisParam, CHAR_BUF_LENGTH, " %s(%s)", mcinputtable[i].name,
             (*mcinputtypes[mcinputtable[i].type].parminfo)
                 (mcinputtable[i].name));
+    if (strlen(Parameters) + strlen(ThisParam) + 1 >= CHAR_BUF_LENGTH) break;
     strcat(Parameters, ThisParam);
-    if (strlen(Parameters) >= CHAR_BUF_LENGTH-64) break;
   }
 
   /* output data ============================================================ */
@@ -1168,6 +1168,7 @@ static void mcruninfo_out(char *pre, FILE *f)
     fprintf(f, "%sNodes: %i\n",        pre, mpi_node_count);
 #endif
 
+  // TODO Consider replacing this by a a call to `mcparameterinfo_out(pre+"Param: ", f)`
   /* output parameter string ================================================ */
   for(i = 0; i < numipar; i++) {
       if (mcinputtable[i].par){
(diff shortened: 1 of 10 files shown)


willend:
I am planning a new McStas release post vacation, hope to get this merged in prior to the related release work.

g5t:
Great! Do you have any ideas about the syntax and whether it is even possible to expose the singular `'/'` as a token in `instrument.l`? I haven't looked back at it, but my failure to find a solution prevented my implementation of (what I think is) a cleaner syntax.

willend:
Let's look at all of this together next week as suggested.

PR #1444, opened 2023-05-10
Mccode 3 fix 1219

### LINKED ISSUE #1219: McXtrace mxplot calls mcplot-pyqtgraph when 'x' is pressed.
There is a hard-coded call of mcplot somewhere, needs a mccode_config based switch

ebknudsen:
This should fix #1219 (although the commit message says otherwise. 
Simply copy the solution from further above where the mc/mx prefix is deduced from the MCCODE environment. variable.


willend:
DOH! Overlooked this one releasing McXtrace 3.2, well never mind... :-)

PR #1454, opened 2023-05-22
Update MCPL_output.comp filename

g5t:
The MCPL_output documentation string indicated that a unique filename would be used if no filename was provided. This was not true, and instead the filename `[output_directory]/.mcpl` would be used. Multiple MCPL_output components would use the same filename if not given a filename value, and either wrote to the same file or overwrote each other depending on implementation details in MCPL. Furthermore, the produced filename is hidden on UNIX-like systems, so a user may never find their output file.

The component has been changed to insert the (unique) component name into the filename string if another value is not supplied. This should produce MCPL output files like `[output_directory]/[component_name].mcpl` which a) are unique and b) are not hidden.

### TODO
If this change is acceptable, the same change is likely necessary in the other MCPL_output components 'noacc' and in `mcxtrace-comps/misc` and could be appended to this pull request.

### DIFF
--- a/mcstas-comps/misc/MCPL_output.comp
+++ b/mcstas-comps/misc/MCPL_output.comp
@@ -37,7 +37,7 @@
 * %P
 * INPUT PARAMETERS
 *
-* filename: [str]         Name of neutron file to write. Default is standard output [string]. If not given, a unique name will be used.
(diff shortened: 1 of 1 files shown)


willend:
Planning to include this in forthcoming McStas release (post vacation)

willend:
Very small but very relevant edit, merging in.

PR #1455, opened 2023-05-23
individual sasmodels and two examples

jorobledo:
Still missing a couple of models. Added two examples for testing in the examples folder.


willend:
Great! :) As this is simply additions of comps / c snippets I will merge and play. 👍

Issue #1461, opened 2023-06-08
Possible memory leak


g5t:
When trying to find the cause of the segmentation fault which led to #1460, I found that in some cases the `mcdetector_statistics` function in `common/lib/share/mccode-r.c` can leak memory.
This memory leak seems to be a result of this allocation:
https://github.com/McStasMcXtrace/McCode/blob/46cb28eade6b685d6a76f1c45efaac156a1b1405/common/lib/share/mccode-r.c#L769

A simple instrument that should show this behaviour is
```
DEFINE INSTRUMENT memory_leak()
TRACE
COMPONENT source = Source_simple(lambda0 = 2, dlambda = 1, target_index = 1) AT (0, 0, 0) ABSOLUTE
COMPONENT monitor = L_monitor(Lmin=0, Lmax=4) AT (0, 0, 1) RELATIVE  PREVIOUS
END
```
if you save it locally, and parse, compile with debugging symbols, and run under Valgrind's Memcheck you will see similar output to the following:
```shell
$ mcstas memory_leak.instr 

-----------------------------------------------------------

Generating single GPU kernel or single CPU section layout: 

-----------------------------------------------------------

Generating GPU/CPU -DFUNNEL layout:

-----------------------------------------------------------
CFLAGS=
$ gcc -g -o memory_leak.out memory_leak.c -lm
$ valgrind --leak-check=full ./memory_leak.out -n 1
==7101== Memcheck, a memory error detector
==7101== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==7101== Using Valgrind-3.20.0 and LibVEX; rerun with -h for copyright info
==7101== Command: ./memory_leak.out -n 1
==7101== 
Instrument parameters for memory_leak (memory_leak.instr)
*** TRACE end *** 
Detector: monitor_I=3.31645 monitor_ERR=3.31645 monitor_N=1 "monitor.dat"
==7101== 
==7101== HEAP SUMMARY:
==7101==     in use at exit: 640 bytes in 1 blocks
==7101==   total heap usage: 23 allocs, 22 frees, 20,377 bytes allocated
==7101== 
==7101== 640 bytes in 1 blocks are definitely lost in loss record 1 of 1
==7101==    at 0x4848464: calloc (vg_replace_malloc.c:1340)
==7101==    by 0x40420E: mcdetector_statistics (memory_leak.c:2061)
==7101==    by 0x4058F7: detector_import (memory_leak.c:2331)
[...]

PR #1462, opened 2023-06-08
[Fix] memory leak from mcdetector_statistics

g5t:
As discussed in #1461 there is a memory leak in `mcdetector_statistics`, and it appears that only the case where `this_p1` is assigned is a problem. This was not very clear when looking at the code because it looks like the triple-`AND` `if` statement before the assignment is  doubly redundant (if `this_p1` has been assigned, the other two statements must be true).

The fix modifies `mcdetector_out_1D` to verify whether the returned `MCDETECTOR` structure has the same `p1` pointer,
and frees the memory allocated in the call to `mcdetector_statistics` if the pointers do not match.
I _think_ the returned `MCDETECTOR` from `mcdetector_out_1D` is not used anywhere (so that the lost 'statistics' calculation result is not missed), but I can not guarantee that to be true.

The function is used (or defined) in 
```shell
./common/lib/share/read_table-lib.c
./common/lib/share/mccode-r.h.in
./common/lib/share/mccode-r.c
./mcstas-comps/examples/BTsimple.instr
./mcstas-comps/share/monitor_nd-lib.c
./mcstas-comps/share/monitor_nd_noacc-lib.c
./mcstas/src/mcformat.c.in
./mcxtrace-comps/share/monitor_nd-lib.c
./mcxtrace/src/mcformat.c.in
```
which should be checked for use of the `MCDETECTOR` returned from `mcdtector_out_1D` before accepting this PR.

### DIFF
--- a/common/lib/share/mccode-r.c
+++ b/common/lib/share/mccode-r.c
@@ -2197,10 +2197,16 @@ MCDETECTOR mcdetector_out_1D(char *t, char *xl, char *yl,
 
 #ifdef USE_NEXUS
   if (strcasestr(detector.format, "NeXus"))
-    return(mcdetector_out_1D_nexus(detector));
+    detector = mcdetector_out_1D_nexus(detector);
   else
 #endif
-    return(mcdetector_out_1D_ascii(detector));
+    detector = mcdetector_out_1D_ascii(detector);
+  if (detector.p1 != p1 && detector.p1) {
+    // mcdetector_statistics allocated memory but it hasn't been freed.
+    free(detector.p1);
+    detector.p1 = p1;
+  }
+  return detector;
 
 } /* mcdetector_out_1D */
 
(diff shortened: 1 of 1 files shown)


g5t:
It seems the fix as-is is not good enough, since any custom component could use `mcdetector_out_1D` directly _**or**_ one of the three wrapping Macros/functions that call it and return its value; namely `Table_Write` in `common/lib/share/read_table-lib.c`, `DETECTOR_OUT_1D` in `common/lib/share/mccode-r.h.in`, and `Monitor_nD_Save` in `*-comps/share/monitor_nd-lib.c`
Importantly, the last _is_ being used already by `mcstas-comps/contrib/StatisticalChopper_Monitor.comp` which performs calculations based on the returned `detector.p1` values.

Would it be acceptable to **copy** the results from `this_p1` to the original `detector.p1`, then free the memory, instead of assigning the pointer? Is the passed-in `detector.p1` used outside of the statistics function via a pointer that is not returned?

g5t:
It _looks_ like the behaviour with the memory leak meant that the `StatisticalChopper_Monitor.comp` already did not work for 1-D monitors, since the (N,) sized data it attempted to analyse was replaced by a (N,4) array.

The new commit extends the fix to include reproducing the 'expected' structure properties for a 1-D array in the 'McCode' format after the file has been written. This allows the 1-D correlation algorithm to work and produce output which _might_ be right.

willend:
Will get this into next McStas release (post vacation)

PR #1463, opened 2023-06-09
fixing what appears to be a typo

bsobhani:


willend:
Thank you for noticing @bsobhani!

PR #1475, opened 2023-07-12
Implement LITERAL keyword

g5t:
This branch includes an implementation fulfilling my request #1470 

Unfortunately, it is on top of the current version of #1438 and includes bug fixes and other developments.
If this is acceptable, I can extract just the new-keyword related changes before merging to keep the repository history cleaner.

PR #1477, opened 2023-07-13
Implement `METADATA` keyword and add instrument parameter units

g5t:
- Replaces #1475 by implementing `METADATA` instead of `LITERAL`.
- Implements `INSTRUMENT` parameters with optional units defined as `{type} name [/ "{unit}"] [= {value}]`.


willend:
Ran succesful test on macOS(clang) and mostly successful on Ubuntu(gcc-13) - merging in.

PR #1478, opened 2023-07-14
[Fix] Metadata issues

g5t:
- Metadata added in the component definition was added to the instances of the component as expected, but only its pointer was copied. This meant that all instances *shared* the same underlying data structure and, e.g., printing the metadata names for one of the components would either show multiple copies of the component-defined metadata or none.
- Instantiating or COPYing an instantiated component now *copies* the metadata list, including copying the 'source', 'name', and 'type' strings. The underlying 'List' of metadata `lines` is still shared between instances, which seems like the correct behavior.

- Printing the list of components which defined metadata had a memory allocation error. This could lead to printing of weird characters.
- The correct amount of memory is now allocated to track 'known' component names, and the printing is moved to a single `printf` call.

- Metadata _should_ provide a MIME type (and not just a file extension) this requires allowing the first element following METADATA to be a a string, since MIME types include a forward slash and may include other non-idenifier-allowed characters.
- The name of the metadata may require non-identifier characters in the future, to allow for more-flexible identification of metadata origins or expected use cases.
- To address these uses, a piece of metadata can now be created using METADATA id id, METADATA string id, METADATA id string, or METADATA string string.
- If the name is specified as a string, it will be recorded with its double quotes and escaped to be a valid C char* style string literal. This requires use of the same escaping to retrieve the metadata from the compiled instrument.

### DIFF
--- a/common/lib/share/mccode-r.c
+++ b/common/lib/share/mccode-r.c
@@ -3950,11 +3950,10 @@ mchelp(char *pgmname)
 "  -h        --help           Show this help message.\n"
 "  -i        --info           Detailed instrument information.\n"
 "  --list-parameters          Print the instrument parameters to standard out\n"
-"  --metadata-list            Print names of components which defined metadata\n"
-"  --metadata-defined COMP{::NAME}\n"
-"                             Print component defined metadata text names\n"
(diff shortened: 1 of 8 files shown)


willend:
Yay! Looks good on the annoying machine also now... ;-)

PR #1481, opened 2023-08-14
Radial collimator bug fixing and feature expansion

Lomholy:
Added possibility of using the radial collimator for focusing, and defining inner and outer yheight separately.


Lomholy:
Hi Peter,

A bug fix might be a little much to say, but when trying to use the
component as a focusing collimator it did not allow neutrons to pass
through at all. This seemed a bug.

I am currently making an instrument - The SALSA instrument at the ILL which
uses these new features. Should this just be uploaded to the example
instruments folder, and referenced in the docstring?

Kind regards,
Daniel

Den man. 14. aug. 2023 kl. 16.19 skrev Peter Willendrup <
***@***.***>:

willend:
Hi Daniel,

Thanks! 👍 

And yes, an addition of `ILL_SALSA` to the examples directory would be absolutely great! I believe you can add the instr to the same PR - but if it feels easier / better to split in two, feel free.

Cheers
Peter

willend:
@Lomholy thanks for that! :-)

I ran a quick test locally, looks nice! You probably developed with / using McStas 3.x? 
(What you didn't know is that master is 2.x and mccode-3 is 3.x - but this is also not too clearly documented, sorry. :-) 
I will eventually fix that, plus ensure that the code ends up working correctly on both code variants.)

Lomholy:
@willend  You are very welcome!

I did develop this with McStas 3.3 i think. For the future I will remember that mccode-3 is 3.x.

willend:
What should eventually happen is that mccode-3 becomes 'main' which is the 'new default' for GitHub... :)

PR #1483, opened 2023-08-15
[Fix] failing mcrun scans due to captured output

### LINKED ISSUE #1482: Scanning via `mcrun -N` can fail due to `STDOUT` capture
## Problem
When run in 'scan' mode, `mcrun` captures `STDOUT` and `STDERR` in order to identify detector lines.
https://github.com/McStasMcXtrace/McCode/blob/c725e6fcd621f1103b37b5d46c913be60df8dc1a/tools/Python/mcrun/optimisation.py#L261

For not-entirely-clear reasons, in some cases the scan can hang indefinitely -- possibly due to overflowing the buffer as seems to be indicated in the [Popen.communicate documentation](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate)

## Solution
This problem is fixed by switching from `STDOUT` parsing to reading the requisite information back from each scan-point `mccode.sim` file, and *not* capturing either `STDOUT` or `STDERR`.

#### Scan progress indication
Output from each simulation step in the scan will be sent to `STDOUT` allowing a user to easily identify that a long-running process is progressing instead of wondering why no output is produced, in contrast to the single-point (no `-N`) mode of `mcrun`. 

#### Possible undefined behavior
In its current state, a user could inadvertently break `mcrun` in scan mode by writing a component that produces lines which match the regular expression used to identify detectors
https://github.com/McStasMcXtrace/McCode/blob/c725e6fcd621f1103b37b5d46c913be60df8dc1a/tools/Python/mcrun/mccode.py#L379

g5t:
- Modifies mcrun/optimisation functions/methods to read-back written detector information from 'mccode.sim' file(s) instead of attempting to store and parse all output produced during the simulation.
- The class responsible for parsing the standard output text, McStasResult, was no longer used, and has been removed.
- The function build_mccodesim_header has been refactored to use the 'intervals' keys directly instead of passing them separately.
- The replacement for McStasResult is a function mcsimdetectors which reads the information back from the mccode.sim file produced in each scan or optimisation step.
- Calling the compiled instrument is now handled by `subprocess.run` instead of `subprocess.Popen`, as is advised by the Python module documentaiton.

If accepted, this PR fixes #1482


willend:
👍 you've been at it again @g5t :-)

One thing I'd like us to test initially is that the "file-less" detector outputs from 0D monitors is still picked up when scanning (I believe it should be, output available in subfolder mccode.sim's)

g5t:
It seems to work. `linup-1.instr` has a `Monitor.comp` named `sng`:
```cmd
mcrun linup-1.instr -n 1000000 -N 51 TTM=-71,-77
```
image

willend:
Testing in progress

willend:
I've run the test suite with this code today, column `McStas_CPU_MPICC_GT_5e7` in https://new-nightly.mcstas.org/2023-08-17/overview.html 

No evident side effects (the failures are from missing NeXus and GSL libs on the machine in question).

Marging in - then we will take any collateral damage as It may appear.

PR #1518, opened 2023-11-07
Update NPI_tof_dhkl_detector.comp

g5t:
Remove an unmatched extraneous closing brace in the `DECLARE` block.


willend:
Very strange that these little things have gone unnoticed... And more so that ESS_BEER seems to run with this inclusion?!?

PR #1521, opened 2023-11-20
Single crystal PG powder fix

mads-bertelsen:
The PG and powder mode for the Single_crystal component rotates the neutron velocities randomly to simulate the random orientation of crystallites. This change of coordinate system was not reverted when propagating the neutron to the scattering position inside of the crystal, and it could thus happen the ray was moved outside of the crystal.

The code was fixed by storing and using the velocity in the component frame and using that just for the propagation when one of the two modes are used.

The Single_crystal_process in Union was checked for the same problem, it was not relevant in that implementation, but other updates the powder system was moved to the Union equivalent, as it now rotates around three directions instead of just two, which is necessary to make a completely random crystal orientation in the powder.

### DIFF
--- a/mcstas-comps/samples/Single_crystal.comp
+++ b/mcstas-comps/samples/Single_crystal.comp
@@ -1396,10 +1396,10 @@ TRACE
         && fabs(kiz - hkl_info.kiz) < deltak) {
         hkl_info.nb_reuses++;
 
-	    /* Restore in case of matching event (e.g. SPLIT) */
-	    coh_refl = hkl_info.coh_refl;
+	/* Restore in case of matching event (e.g. SPLIT) */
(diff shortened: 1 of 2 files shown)


willend:
I ran a few tests with Test_Monochromators and Test_SX and all looks good. 
Did a reset of some whitespace to original structure.

willend (approved):
I believe the changes are OK, but have pushed a few reverts of modified whitespace. (This in my point of view always clutters what the meaning/implication of the "actual" fix is. :-) )

farhi:
This issue fix has created bug #1732
So the fix is not relevant, and it is probably better to keep some escaping neutrons, rather than having no powder mode.

PR #1522, opened 2023-11-27
Adding of a contributed component, and corresponding test instrument

Lomholy:
This pull request adds a new bent monochromator which follows roughly the same algorithm as Jan Saroun in doi:10.1016/j.nima.2004.04.197

The test instrument accompanying the monchromator component is capable of recreating the rocking curve presented in above paper. A comparison of the two was made and produced this graph which I am trying to attach:
Rocking_curve_simres_mcstas_compare


willend:
I have given some feedback to @Lomholy via email and we agreed that he will bring a fresh PR after some modifications. 👍

PR #1533, opened 2023-12-07
Small fixes in monitors

mads-bertelsen:
Monitor_4PI
Wrong use of NAME_CURRENT_COMPONENT in save section fixed

TOFlog_monitor
Used static memory, moved to dynamic and removed static limit on memory

PSD_monitor_psf
Still had state and polarization parameters that would not compile

### DIFF
--- a/mcstas-comps/monitors/Monitor_4PI.comp
+++ b/mcstas-comps/monitors/Monitor_4PI.comp
@@ -71,7 +71,9 @@ TRACE
 
 SAVE
   %{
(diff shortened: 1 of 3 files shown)


willend (approved):
Looks good, all meaningful changes. I added two small commits removing the for-loops with 0-init, no longer needed with the use of create_darr*() routines. 

(We should perhaps check if there are other monitors around that still perform that step.)

willend:
I've removed remaining for-loop init's in https://github.com/McStasMcXtrace/McCode/commit/2a52e5848588cb29daa69db85346d7bf32748efe. Now running integration test.

PR #1549, opened 2024-01-22
Fix of Phonon_simple weight factor

mads-bertelsen:
Fixed error in weigth factor calculation of phonon simple, decreasing the intensity significantly.

The example instrument has been rerun with the new version and the decrease is as expected.

The component is also used in RITA-II, though without any example for testing.


willend:
This all looks good, thank you @mads-bertelsen, merging to main now.

PR #1550, opened 2024-01-23
Adding of Monochromator_bent.comp

Lomholy:
This pull adds a component that simulates a bent perfect crystal acting as a monochromator. It also adds an instrument that tests this component, the Test_monochromator_bent.instr, which should produce a rocking curve like the following:
mcstas_simres_comparison
When rocked back and forth around its nominel Bragg angle, by the listed arc minutes.

Also updates the SALSA instrument to use the new monochromator, in the same setup as the actual SALSA instrument has.

### DIFF
--- a/mcstas-comps/contrib/Monochromator_bent.comp
+++ b/mcstas-comps/contrib/Monochromator_bent.comp
@@ -1,0 +1,905 @@
+/*******************************************************************************
+*
+* McStas, neutron ray-tracing package
+*         Copyright 1997-2002, All rights reserved
+*         Risoe National Laboratory, Roskilde, Denmark
+*         Institut Laue Langevin, Grenoble, France
+*
+* Component: Monochromator
+*
+* %I
+* Written by: Daniel Lomholt Christensen <dlc@math.ku.dk>
+* Based on the model implemented by Jan Šaroun in the paper published in
+* Nuclear Instruments and Methods in Physics Research Section A: Accelerators, Spectrometers, Detectors and Associated Equipment
+* Volume 529, Issues 1 through 3, 21 August 2004, Pages 162-165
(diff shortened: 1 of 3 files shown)


willend:
Hi @Lomholy and thanks for the PR,

I've tentatively added another input parameter `mono_rock` in the attempt of allowing to do a rocking curve... Do you agree with the principle of this approach? I.e. using `mcrun` one should run something like 

`mcrun -c ILL_SALSA.instr mono_rock=-4,4 -N17 -n1e6 --mpi=auto --autoplot`

This could for instance be added as an Example line.

(In the process I also came across https://github.com/McStasMcXtrace/McCode/issues/1551 - workaround in the `_Phic` 0D monitor also added in ILL_SALSA)

willend:
Oops, seems I made the above, suggest changes on main rather than your PR. Sorry about that. Will get that fixed asap.

Lomholy:
Hi @willend 

I do agree with that approach, if that `mono_rock` variable is added to the rotation of the monochromator, and the adding of the example line in the instrument documentation.

Lomholy:
Oops, it seems I was also a little too fast in responding. 

I don't think I agree with adding the above mentioned changes on the `ILL_SALSA.instr`, as this one does not usually use a rocking curve, but I do agree with adding the above mentioned changes on the `Test_Monochromator.instr`.

Kind regards, 
Daniel

willend:
Sorry @Lomholy!

 Apparently working on the `main` version of the file I didn't realise that your new `mono_rotation` already allows to directly perform a rocking curve scan... :)

willend:
I will simply merge 'as is' for now - then work on the rocking curve idea on the simpler Test instrument later.

Issue #1557, opened 2024-01-31
No extension is added by `mcfull_file` for filenames including '.'


g5t:
https://github.com/McStasMcXtrace/McCode/blob/60ab0032c982ea8e58d5c0b026beb22e0a42bdce/mcstas-comps/misc/MCPL_output.comp#L107

## Background
If an instrument uses a parameter to specify the output filename of an `MCPL_output` component, a user may specify filename that includes one or more '.' characters, e.g., as part of an absolute path or because they expect to produce a `.mcpl` file.

`mcfull_file(name, extension)`, nominally, prepends the output directory to the user/component provided `name` followed by a '.' and finally `extension`.
To avoid mixing absolute and in-output-directory file name specifications, `mcfull_file` only prepends the output directory if no path-separator characters are present in `name`.
It also does not append the '.' or `extension` if `name` contains any '.' characters (or `extension` is `NULL`).
These are likely sensible choices in most cases.

## Problem
This behavior is problematic in the case of a user-specified filename including one or more '.' characters  _while_ running under MPI, since each process uses `mcfull_file` with a per-process extension to avoid simultaneously writing to a single file.
The head process can concatenate all MCPL files by rebuilding their individual file names, again using `mcfull_file`.

Since no extension is added to the filename, all processes use the same file.
All processes write to that file and an error is only raised at the merge step -- after the full simulation has finished.

## Solution
Do not use `mcfull_file` to append the special per-process extension. Instead, use standard C `char` array manipulations to build the filename for each process.

PR #1558, opened 2024-01-31
[Fix] MPI MCPL_output filename error

g5t:
This is my proposed fix for #1557


willend:
I ran the patch on a couple of machines - and it seems to work OK...

(As a side-remark a crash on macOS/Silicon/mamba with "bleeding edge" openmpi=5.0.1 has animated me to specify openmpi=4 in environment.yml for now as pr. https://github.com/McStasMcXtrace/McCode/commit/d39588b87f6ba16668eedbfe4480cbc31f510892)

PR #1560, opened 2024-02-02
Interp regular fix

sq-meng:
- Comparator function fixed to work with floating point values correctly.
- interpolator number of bins fixed - was off by one.
- `vector` array freed after use, for good measure.
- `round()` instead of `floor()` used when storing and retrieving values in `regular` interpolation.

### DIFF
--- a/common/lib/share/interpolation-lib.c
+++ b/common/lib/share/interpolation-lib.c
@@ -257,11 +257,11 @@ vertex* kdtree_nearestNeighbour(vertex* v, treeNode *tree) {
 /******************************************************************************/
 /* interpolator_double_vector_compare: comparator for double qsort */
 int interpolator_double_vector_compare(void const *a, void const *b) {
-  if ((*(double*)a - *(double*)b) > 0 )
+  if (*(double*)a > *(double*)b)
   {
     return 1;
   }
-  else if ((*(double*)a - *(double*)b) < 0 )
+  else if (*(double*)a < *(double*)b)
   {
     return -1;
   }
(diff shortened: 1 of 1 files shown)


willend:
Thanks for the PR @sq-meng :-) I might take a little while examining, testing etc. - but will try to get it done ~ within a weeks time or so.

willend:
Since last I've fallen ill with something flu-like, so still didn't have energy to look @sq-meng... But I did not forget you...

sq-meng:
@willend Sorry to hear that, get well soon!

willend:
@sq-meng I am back at reduced capacity and have managed to run a couple of hand-held tests:
http://tmp.mcstas.org/oneoff_2024-02-08_output.html 
Left column was run on GPU's, right column on CPU.

CPU is fine, potential minor issues on GPU:
* The issues of instruments that don't compile are **unrelated** to this PR
* The "red" cells for the Union_ instruments are **unrelated** to this PR
* Leaving only the Test_Pol_Tabled cell which is off on GPU - and which **does** use interpolation-lib

Anyway, this is quite acceptable and on the scale of "I will come back to that later myself" :-) 

Thanks again for contributing - I am merging in the PR

willend:
For sake of clarity, I will open an issue to look in to the implementation details of interpolation-lib on GPU. Here are overview plots from both simulated cases on

CPU:
Screenshot 2024-02-09 at 08 54 18
Screenshot 2024-02-09 at 08 54 32

GPU:
Screenshot 2024-02-09 at 08 53 54
Screenshot 2024-02-09 at 08 54 07

- so irrespective of simulated case something is off... :-) 

```
* %Example: -n1e4 MF=flipfield.dat Detector: pol_6_I=1.0
* %Example: -n1e4 MF=constfield.dat Detector: polx_6_I=0.05
*
* %Parameters
* MF:              [string] Field definition file
* zdepth:               [m] Z-dimension of field
* interpol_method: [string] Choice of interpolation method "kdtree" (default on CPU) / "regular" (default on GPU)
```

PR #1593, opened 2024-03-21
[Add] metadata-name accessor, [Resolve] issue #1592

### LINKED ISSUE #1592: Possible metadata runtime issue
The `metadata_table_defined` method is not used in the McStas/McXtrace runtime; but it _could_ be used in a component to check if a specified `{component name | component instance name}:{metadata name}` key exists in the global `metadata_table`.
It splits the provided key at the `:` (if it exists) and returns the number of entries in the table with matching part(s) -- if the key has no `:`, the key is interpreted as the `source` and the returned number is the number of `METADATA` entries provided from the `source` component definition or component instance specification.

The actual implementation avoids accessing a `NULL` key
https://github.com/McStasMcXtrace/McCode/blob/7859d04849c7dbc0336e37001be6ceb65cea889a/common/lib/share/metadata-r.c#L34-L37

Since the combination `(source, NULL)` returns the number of `source`-defined entries, some users _might_ expect that `(NULL, NULL)` _should_ return the total number of defined entries.

Others might instead expect the shortcut case to return `0`, since _no_ entries match.

---
## Desired outcome
A decision for which behavior is intended should be made. Following this, the function return should be better documented to aid component definition authors who may like to access stored METADATA.

g5t:
Resolves #1592 by changing the return value of `metadata_table_defined` to `0` if the provided key is empty _and_ adding documentation for the metadata accessor methods.

Adds a method that returns the first-defined metadata _name_ for a known _source_, e.g., the component instance name.
This is mainly a convenience function for users of [a new component](https://github.com/g5t/mccode-file) that can, e.g., make an `instr` file using OFF geometry component(s) self-contained.

### DIFF
--- a/common/lib/share/mccode-r.h.in
+++ b/common/lib/share/mccode-r.h.in
@@ -261,6 +261,7 @@ typedef struct metadata_table_struct metadata_table_t;
 char * metadata_table_key_component(char* key);
 char * metadata_table_key_literal(char * key);
 int metadata_table_defined(int, metadata_table_t *, char *);
+char * metadata_table_name(int, metadata_table_t *, char *);
 char * metadata_table_type(int, metadata_table_t *, char *);
 char * metadata_table_literal(int, metadata_table_t *, char *);
 void metadata_table_print_all_keys(int no, metadata_table_t * tab);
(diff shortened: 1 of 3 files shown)


willend:
As discussed I am merging in the metadata-accessor code

Issue #1595, opened 2024-04-03
C macro version awareness


g5t:
Version information _in generated C code_ could be useful to component authors wishing to make use of bleeding-edge developments without breaking backwards compatibility for older McCode versions. One way to provide this would be through an integer-valued macro, e.g., `__MCCODE_VERSION__`, created via something like 

```c
#define __MCCODE_VERSION__ 304038L
```

Then a component block that is C code can use, e.g.,
```c
#if defined(__MCCODE_VERSION__) && __MCCODE_VERSION__ >= 304003L
// do something only possible thanks to a patch in 3.4.3
#else
// use a workaround or issue a warning/error
#endif
```

g5t:
This was addressed by #1596

g5t:
This was a duplicate of #1215

PR #1596, opened 2024-04-03
[Add] C macro integer version

g5t:
Implementing the fix proposed in #1595 (and mistakenly including part of #1593)

The CMake project version information is used to build an integer valued C macro, with the exact value inserted at configure time.

This could lead to a disconnect between package version and built version due to the use of the `MCVERSION` variable.
@willend How would you prefer this be resolved?


g5t:
The error has been softened to a warning, with a default version 3.99.999 used in case `MCVERSION` is not set.

PR #1599, opened 2024-04-16
Correction to allow multiple comps (e.g. Monochromators) requesting USERVARS

willend:
Also add row/col based USERVARS in Monochromator_curved


willend:
Tested out OK on willend/McCode -> merging

PR #1603, opened 2024-05-02
Let code-generated class_{$COMP}_trace functions become void type

willend:
The existing McCode 3 generated comp-functions all end in `return(_comp);` 

This is meaningful in e.g. `INITIALIZE` and `SAVE`, less so in `TRACE` - where the current implementation currently does not make any use of such a returned `_comp`.

This further meant that the in-component `ABSORB` keyword could in fact not contain `return;` - having the side-effect that some components would continue performing calculations on an already "absorbed" component...

The edits here generates trace-functions ending in simple `return;` statements and adds back `return;` in the comp trace `ABSORB` macro.


willend:
Tests have already run successfully on my own fork and produced functional installers, see e.g. https://github.com/willend/McCode/actions/runs/8916461017

PR #1605, opened 2024-05-09
Workarounds/solutions for systematic slowdown when using Conda-provided compilers 

willend:
Using a Ubuntu 22.04 system it was recently discovered that simulations based on identical c-code would systematically become ~ 30% slower when using `${CONDA_PREFIX}/bin/x86_64-conda-linux-gnu-cc` than `/usr/bin/gcc-12` (both gcc-12.3)

Fig 1: Relative runtime of single-core 1e7 runs of the McStas example suite, conda-gcc vs system-gcc. Shows that almost all instruments were substantially slowed down.
[Relative_runtime.pdf](https://github.com/McStasMcXtrace/McCode/files/15258669/Relative_runtime.pdf)

Fig 2: Histogrammed version of the Fig 1 data, showing a mean slowdown of  ~ 30% 
[distribution.pdf](https://github.com/McStasMcXtrace/McCode/files/15258671/distribution.pdf)

Fig 3: Investigating further, the slowdown scaled linearly with problem size, here investigated for the case `mini.instr`
[mini-1e3-1e11.pdf](https://github.com/McStasMcXtrace/McCode/files/15258771/mini-1e3-1e11.pdf)

Following an thorough, detailed investigation (thanks for ping-ponging on this @tkittel) it was found that:
1. Adding `CFLAGS` of `-fno-PIC -fPIE -flto -mtune=native -march=native -fno-math-errno -ftree-vectorize` helped 
2. Conda (for sake for portability) ships an **ancient** `libc` and `math.h` which lead to a costly overhead of not directly using e.g. `__builtin_isnan()` -> the solution being to redefine use of `isnan` and `isinf` to the internally provided functions when using `gcc` or `clang`
3. Optimising to `-O3` seems to not cost much (~10% longer compilation time) but cutting ~25-50% runtime when added alongside the above points

(An initial glance at the results can be seen at http://tmp.mcstas.org/2024-compiler-comparison.html - "result" columns are `mcstas-3.x-dev_Linux` == conda-compiler with the above patches, `conda-sysgcc_Linux_1e7` == system-gcc run from the otherwise same conda environment. Further explanation to be added later. :-) )

See also https://github.com/mctools/simplebuild/issues/73

@tkittel you are very welcome to review/comment further?


tkittel:
The diff looks great to me! I am curious why there is no CI run, I thought we had fixed that. Or perhaps we have not enabled it for pull requests?

tkittel (commented):
Looks good!

tkittel (approved):

willend:
@tkittel thanks! I think a PR switch is perhaps still missing in the CI. Will check tomorrow, for now I will take my chances and merge ;-)

PR #1607, opened 2024-05-10
Less use of isnan() and isinf() in trace-fcts + ABSORB without return in OPENACC mainloop.

willend:


willend:
All checks completed successfully, merging.

PR #1623, opened 2024-06-06
More work on visualisation 

willend:
Further work on visualisation, new prototypes for cylinder/box and fallback to legacy function types in case of trace==1.

Consequence edits to comps in McStas and McXtrace


willend:
Tests completed successfully. merging.

PR #1635, opened 2024-07-04
Webgl upgrade 2

theogutt:
New react based McDisplay-WebGL frontend launched from mcgui.

Added 2D views(PyQtGraph), that will be further developed

Light mode, dark mode


willend:
@theogutt tests checked out and I have also built locally via merge to local branch (https://github.com/willend/McCode/tree/test-branch-3Dvis)

I guess everything you did for the new server-oriented gui is not here yet? (I still get the semi-classic interface here).

Will merge anyhow and we can look at remaining details tomorrow.

PR #1637, opened 2024-07-06
Add node.js as dependency in mcdisplay-webgl

willend:
mcdisplay-webgl now depends on node.js. 
npm-based dependencies are added at configuration-time via package.json

(Windows required a little special attention, CMake is only able to run npm via npm.cmd)


willend:
A tests checks out, merging

PR #1639, opened 2024-07-11
Leave "classic-mode" webgl implementation in place for potential use on Debian-class systems

willend:
As a fallback solution I am leaving a "webgl-classic" package in place alongside the current one which requires node.js >=18.

Debian stable (and others?) plus Ubuntu 22.04 and earlier provide a default node.js of version 12 which is too old for the current webgl+vite server node.js implementation.

The "classic" files here still support the new added "solid" primitives, but not all the fancy possibilities of the new interface.


willend:
PR tests concluded OK, merging

farhi:
:+1:

PR #1640, opened 2024-07-11
Use install( DIRECTORY ... ) rather than globbing the comp files as pr. email discussion with @farhi

willend:
Should make our install more standard and the ongoing Debian-work a bit easier.


willend:
All tests completed successfully, merging.

willend:
@farhi FYI

PR #1643, opened 2024-07-19
Union example cleanup

mads-bertelsen:
Ongoing work on clean up of examples using Union components. These are now better organized in:

- Tests_union : Small test instruments
- Union_demos : Larger instruments showing the capabilities
- Union_validation : Validation against classic McStas sample components (that were used as basis for Union processes)
- Union_sample_environments : Sample environments meant to be %include 'ed in other files + usage example

This is still work in progress, though starting to take shape. This pull request is added so anyone interested can easily comment if anything is missing / should be changed in the final stages.


willend:
Looks good to me, only a minor merge-issue on the powder validation instrument it seems. (+ failure on macOS 11 which I believe is no longer supported as GitHub runner - I think it is even disabled on current main.) Will proceed with proper review and merge when back from vacation ~ tuesday.

willend:
Good to have this in place @mads-bertelsen, merging.

PR #1649, opened 2024-07-26
Functional Debian build

farhi:
Here is a list of the modifications:
- removed 'launchers' `postinst` section (was never executed as this directory is not any more in resources)
- add `cif2hkl` as a dependency, removed from build as it is now a proper Debian package
- attempt to remove `-environment` script from BINDIR (does not seem to fully work, but no side effect)
- fixed Debian build for tools
- fixed annoying `postinst` linking of all `/usr/bin/*` to themselves (nasty)
- use default Debian directory for build. The side effect is that we can only install one version of mcxtrace/mcstas, in sync with Debian policy.

This PR has been tested in a VM for:
- installation
- calculation of e.g. `Test_PowderN`
- plotting of results
- uninstallation

That's all folks !


willend:
Sounds good, will test when I get back from vacation.

About the -environment script, please keep in mind that this is what is used for mc/xgui File->Open environment shell, perhaps we should have a mccode_config entry point for the location, which could then differ on Debian wrt. other platforms?

(I myself find it handy having it on the path on macOS, but this anyway mostly happens via a manually created symlink.)

willend:
Also, having only one McStas / McXtrace available on the system is OK IMHO - in line with what most users will have anyway and conda behaves exactly the same.

farhi:
New tests are OK. I think we're all set !
I also took the opportunity to fix some more tools errors, and handle Debian bug:
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076507

Best for the Debian build would be to get rid of the postinst/postrm (or at least clean them up). 
This implies that e.g. the documentation should be generated at package build, not at install.

willend:
This sounds very good!

Getting rid of postinst/prerm should be perfectly OK - and non-Debian systems have their own solutions for doc generation:

* Windows “legacy” batch-postinstall takes care
* macOS “legacy”, packaged build-time
* conda all platforms: initiated by first mc/xrun finding no index.html in $MCCODE (used has write access)
* RPM - I think we should stop generating these. Use conda.

willend:
Also, as communicated via e-mail, the nightlies have been reinstated, ran on current main this afternoon and all looks normal.

willend:
Merge and fingers crossed - 🤞- if anything fails this can be resolved next week.

farhi:
Thanks, that good work indeed :+1: . BTW, I fixed many oddities that were remaining in the plotters plot/display. 
Next step is the doc generation at build, and add some man pages as MD files, probably in the `doc` directory...

PR #1652, opened 2024-07-30
Enable legacy folder structure in Windows cross-compile "monolithic" builds

willend:
As per discussion with @farhi

### DIFF
--- a/buildscripts/build_windows_mcstas
+++ b/buildscripts/build_windows_mcstas
@@ -34,6 +34,9 @@ WORK=`pwd`
 rsync -avz --delete --exclude=".svn/"  mcstas-comps/ mcstas-comps-win
 find mcstas-comps-win -type f -not -name \*.off -exec unix2dos \{\} \;
 
+# For now we need to use legacy file destinations on Windows
+export MCCODE_USE_LEGACY_DESTINATIONS=ON
+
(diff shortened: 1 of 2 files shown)


willend:
Builds seem mostly functional with recent path edits + setting LEGACY paths. 

But I have found an issue with recent mcplot edits @farhi :
The default (conda-based) version of pyqtgraph is 0.13.7 - which looks to be incompatible with your recent fixes.... 
I can not click to get the single plot from the gui - so likely some sort of switch-based solution needs to be put in place.
(What was there until a few days ago did work on the conda-platforms)

For now reproduced on Windows, but I would think it behaves the same on the other conda-platforms... Will keep you posted.

willend:
@farhi the above is also broken on a Ubuntu running in a conda environment...

willend:
@farhi it turns out the issue is that conda-forge is still PyQt5 only - and your recent edits for the modifier handling seems to work for PyQt6 only.

willend:
@farhi yup, coming in via https://github.com/McStasMcXtrace/McCode/pull/1655

willend:
Ah ok may affect mcdisplay also, will test later. https://github.com/McStasMcXtrace/McCode/pull/1655 lets mcplot function

PR #1654, opened 2024-07-30
Fixes to env script to work properly in both LEGACY and non-LEGACY

willend:
The resulting $MCCODE is correct now, both in LEGACY and non-LEGACY mode


willend:
Checks out OK, merging

Issue #1582, opened 2024-03-05
Plot: PyQtGraph: click on single monitor plots fails


farhi:
When running a simulation and plotting results, a click on any part of the PyQtGraph window brings an error:
```python
Traceback (most recent call last):
  File "/usr/share/mcxtrace/3.4-20240304/tools/Python/mxplot/pyqtgraph/../../mccodelib/pqtgfrontend.py", line 388, in <lambda>
    lambda event: click_handler(
                  ^^^^^^^^^^^^^^
  File "/usr/share/mcxtrace/3.4-20240304/tools/Python/mxplot/pyqtgraph/../../mccodelib/pqtgfrontend.py", line 367, in click_handler
    if int(event.modifiers()) != mod:
       ^^^^^^^^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'KeyboardModifier'
```

willend:
I am afraid the key-modifier stuff needs to be in place courtesy of Windows and macOS - the last commit here renders clicking on the main ui not functional on these systems....

PR #1655, opened 2024-07-30
Different key modifier solutions for PyQt5 and PyQt6

willend:
Fixes the issue with mcplot-pyqtgraph mentioned also in https://github.com/McStasMcXtrace/McCode/pull/1652

### DIFF
--- a/tools/Python/mccodelib/pqtgfrontend.py
+++ b/tools/Python/mccodelib/pqtgfrontend.py
@@ -364,14 +364,24 @@ class McPyqtgraphPlotter():
                 print("click modifier: %s" % str(event.modifiers()))
 
             # prevent action for modifiers mismatch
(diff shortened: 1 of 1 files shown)


farhi:
Thanks for this, this is the right file, as seen from the history there were a few edits, and your solution seems to cover the Qt versions as needed:
- https://github.com/McStasMcXtrace/McCode/commits/main/tools/Python/mccodelib/pqtgfrontend.py

This is an evolution to #1582

willend:
Merging

willend:
Hmm. This was apparently a partial fix only, Qt version correctly detected on Linux but not macOS and Windows.... Looks like I will use a different if(): else: clause

willend:
https://github.com/McStasMcXtrace/McCode/pull/1656 seems to be a full solution

PR #1656, opened 2024-07-31
More elaborate logic in detection of Qt5/6 (and other Qt properites)

willend:
Tested on:
* Ubuntu with Qt6
* Same ubuntu with conda (meaning Qt5)
* macOS Silicon conda
* Windows 11 conda


willend:
Merging

PR #1657, opened 2024-07-31
Clean up root and tool folders

willend:


willend:
Merge cleanup commit (may break links e.g. on mcstas.org and mcxtrace.org - extent to be seen)

PR #1658, opened 2024-07-31
Changes required to let mcdoc generate output in "docdir"

willend:
This PR 
* Ensures packaging of PDF docs also when building "pure CMake", e.g. were missing in conda!
* Lets mcdoc generate output html in the relevant doc folder -> use gnu-like location of docs


willend:
Tests pass - further testing needed on all platforms post to merge

farhi:
Hi,
it seems there is a side effect when running mcdoc: all the comps/instr are moved to the DOC location when writing the HTML. That is strange...

farhi:
OK, the issue seems to be on the  pure Debian side... Inquiring.

farhi:
Solved. I think I have a clean Deb build now. All at the right place. Will push to Debian CI/CD tomorrow. BTW, we have also used the conda tests_pkg script to initiate tests to run right after the build. It executes in a clean chroot with freshly installed packages.

willend:
@farhi great! Were you able to manage entirely on the Debian/salsa side?

farhi:
Yes, all tests and upstream update are on the way.  Now fetching the latest McCode git HEAD for inclusion in salsa.
Tests are much simpler on the Debian side than on the conda side. Just write a bash file with commands inside. The definition of the environment is set from the debian/control files, so all is automatic.

PR #1661, opened 2024-08-01
Fix gcc-14 forthcoming error, and a few compiler warnings

farhi:
Fixes:
- sprintf warnings (possible memory overwrite) at mcstas/mcxtrace compile.
- gcc-14 pointer type mismatch that would have lead to error soon

BTW, the two pygen.c.in are now equal except for one comment line in the header. Perhaps merge and link ?


willend:
@farhi good idea to standardise, let's maybe make an issue to remind us?

@g5t has had similar ideas wrt. cogen.c.in -> a common cogen.c and separate (same name?) for the few defines/text differing between McStas and McXtrace.

PR #1664, opened 2024-08-01
deb/rpm: rename suite-python-ng to suite-python

farhi:
I suggest to remove the '-ng' labelling of deb/rpm meta packages. Only the python tools are left, and only the 'new' 3.x release should be out from now on...
These edits go in this direction.


farhi:
It seems the failing tests are caused by a network failure...

willend:
I am rerunning now - only 1 out of 16 is missing.

willend:
All checks finalised OK now, merging.

Issue #1488, opened 2023-08-29
tools: mcplot/mxplot: allow save as png/tif/eps/svg/etc plots from command line


farhi:
Currently, only the `mcplot-matplotlib` allows to save the plots from the command line (--format).
The `mcplot-svg` does not create an SVG, but displays/creates an HTML plot, and only for a single monitor file

So I'd like to bring-up the possibility to:
- add `--format` to all plotters
- perhaps rename the `mcplot-svg` to `mcplot-html` ? Add a loop to generate HTML's for all monitors when given a directory ?

willend:
@farhi agreed. As far as I remember, the pyqtgraph plotter can only create a "hardcopy" once a plot window exists, so it is really Qt/display oriented... 

And perfectly fine with a rename of mcplot-svg ->  mcplot-html. Perhaps also an extension with an "overview" plot containing iframes with the individual monitor subplots?

( BTW. the `Scipp` (https://scipp.github.io) guys have developed the `plopp`(https://scipp.github.io/plopp/index.html) plotting tool which is supposed to work on the terminal, in Jupyter and potentially in e.g. Qt apps. I was thinking we should have a closer look at some point. )

farhi:
I think I can not do more. PyQTGraph requires indeed a display. The mcplot-svg could be renamed, but for now I think this is fine.

PR #1665, opened 2024-08-02
mcplot-svg -> mcplot-html 

farhi:
This PR proposes:
- a renaming of mcplot-svg -> mcplot-html (as it displays in a browser). This name corres[ponds to what the user may expect (#1488).
-  fix warnings for some `char[10]` allocation that may be too small for  `_%li` (lintian)


willend:
Thanks @farhi, merging

willend:
Oh and a small P.s.: 
I think I saw this tool rendering an empty browser the other day, unsure in which setting -perhaps the old story about Firefox snap...

PR #1666, opened 2024-08-02
DO NOT MERGE  YET (Trace geometry stored in files)

willend:
Hi @farhi, @mads-bertelsen and @g5t,

I am putting together this PR which is a semi-functional implementation for
* Storing component positions and rotations as datasets into NeXus output files
* A first implementation of storing `Monitor_nD` "bin arrays" as datasets into NeXus output files

I am currently stuck with a segfault that happens reproducibly in / around the new `mccode-r` function `mccomp_placement_nexus` e.g. in a run of
`mcrun ILL_H16_IN5_Mantid.instr lambda=4.5 --mpi=auto -s1000 --format=NeXus`

In the PR there is lots of small edits taking into account suggestions from `cppcheck` and `valgrind` - and the crash seems to occur during / post `NXcompmakedata64()` for the `krumm14` comp position...

I've now stared myself blind and hacked around enough that I feel I need input.... 

Does any of you spot an immediate stupid mistake?


willend:
Hi all 
(I +1 @tkittel since you were nice enough to offer help... ;-) )

I am starting to have a better hint at what is going on... 

I "restarted" with a simplified set of edits on https://github.com/willend/McCode/tree/nexus-take2 - in essence this branch only
* Adds a `components` datastructure in the HDF below the upper `entry1/instrument` 
* Each component adds its `name` as an `NXcomponent` under components 
* Two`NXdata` entries are added pr comp position and rotation

The attached test-instruments contain only Arm's - one has 24 total and the other 25 total...
I attach both the instruments and their generated c-codes here:

If you download these pairs of code and ensure to have .c newer than .instr, you should be able to run with e.g. McStas 3.4 or latest version from conda, simply e.g.:

`mcrun --format=NeXus synts24.instr`
(no `-c` please, use what I generated)

[synts24.instr.txt](https://github.com/user-attachments/files/16661665/synts24.instr.txt)
[synts24.c.txt](https://github.com/user-attachments/files/16661666/synts24.c.txt)

[synts25.instr.txt](https://github.com/user-attachments/files/16661668/synts25.instr.txt)
[synts25.c.txt](https://github.com/user-attachments/files/16661670/synts25.c.txt)

On my Ubuntu 24.04 machine (and other Linux's) with gcc I consistently get a segfault when component number 25 tries to open / write its `NXdata` entries for Position and Rotation, more specifically:

```
# McStas 3.x-dev - Aug. 19, 2024: [pid 366791] Signal 11 detected SIGSEGV (Mem Error)
# Simulation: template_simple (synts25.instr) 
# Breakpoint: [_origin_24_setpos] component origin_24=Arm() SETTING [Arm:0] 0.00 % (       0.0/ 1000000.0)
# Date:      Mon Aug 19 16:44:04 2024
# Started:   Mon Aug 19 16:44:04 2024
# Last I/O Error: No such file or directory
# McStas 3.x-dev - Aug. 19, 2024: Simulation stop (abort).
Quit (core dumped)
```

gdb / GEF points to the NXmakedata call(s) from component 25 onwards... 

Any clues / hints / ideas much appreciated!!!!

P

tkittel:
Hi @willend . I confim the issue, including after recompiling with `-O0 -g` and running under gdb on ubuntu 24.04:

```
 Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7fb397a in NX5getattr () from /home/tkittel/scr/miniforge3/envs/mcstasbugaug19/lib/libNeXus.so.1
(gdb) bt
#0  0x00007ffff7fb397a in NX5getattr () from /home/tkittel/scr/miniforge3/envs/mcstasbugaug19/lib/libNeXus.so.1
#1  0x00007ffff7fa8f62 in nxiopendata_ () from /home/tkittel/scr/miniforge3/envs/mcstasbugaug19/lib/libNeXus.so.1
#2  0x000055555556b85d in _origin_24_setpos () at ./synts25.c:8918
#3  0x000055555556bbc5 in init () at ./synts25.c:9004
#4  0x000055555556d717 in mccode_main (argc=6, argv=0x7fffffffbb18) at ./synts25.c:10137
#5  0x000055555555d7f7 in main (argc=6, argv=0x7fffffffbb18) at ./synts25.c:6546
```

Next step sounds horrible: Have the nexus code locally and recompile with debug symbols and `-O0`...

tkittel:
Is it here the code is? 

https://gitlab.psi.ch/sinqdev/sics/-/blob/1f9dd8da95cf434f49bf0bb99c67a97968ec12f3/napi5.c#L2026

It is hard to debug without actually having a local build with debug info enabled, but I do see some potential buffer overflows in that function (however, they are not very likely to be the cause... perhaps).

I tried to build nexus locally, and install it into my conda environment. One issue here is that I want to *NOT* have nexus installed already. So happily I though "I just remove mcstas + nexus pkgs and install mcstas-core". However, it seems mcstas-core has gotten the dependency on nexus. @willend could we perhaps move as many of these optional dependencies away from mcstas-core as possible, and onto mcstas? We can discuss tomorrow, perhaps there are complications I am not aware of.

willend:
Hi @tkittel and thanks for looking at this!

My initial thinking is the problem  could/would likely be related to the McStas (mis?) use of the NeXus c-api... 

At least I think we should initially ensure the issue is not on our side somehow...

Another related problem is that in principle I believe the NeXus team consider the NeXus c-api deprecated...

But yes, let us pull NeXus from the core deps and try building ourselves from 
https://github.com/nexusformat/code

P

tkittel:
I managed to proceed, here is a bit more interesting stack trace:

```
synts25.out: /home/tkittel/work/mcstas-bug-aug19/nexus-4.4.3/src/napi5.c:88: NXI5assert: Assertion `fid != ((void *)0)' failed.

Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
warning: 44	./nptl/pthread_kill.c: No such file or directory
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7c4526e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7c288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff7c2881b in __assert_fail_base (fmt=0x7ffff7dd01e8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x7ffff7fb61fe "fid != ((void *)0)", 
    file=file@entry=0x7ffff7fb6650 "/home/tkittel/work/mcstas-bug-aug19/nexus-4.4.3/src/napi5.c", line=line@entry=88, 
    function=function@entry=0x7ffff7fb71d8 <__PRETTY_FUNCTION__.0> "NXI5assert") at ./assert/assert.c:94
#6  0x00007ffff7c3b507 in __assert_fail (assertion=assertion@entry=0x7ffff7fb61fe "fid != ((void *)0)", 
    file=file@entry=0x7ffff7fb6650 "/home/tkittel/work/mcstas-bug-aug19/nexus-4.4.3/src/napi5.c", line=line@entry=88, 
    function=function@entry=0x7ffff7fb71d8 <__PRETTY_FUNCTION__.0> "NXI5assert") at ./assert/assert.c:103
#7  0x00007ffff7fb1cdc in NXI5assert (fid=<optimized out>) at /home/tkittel/work/mcstas-bug-aug19/nexus-4.4.3/src/napi5.c:88
#8  NX5getattr (fid=<optimized out>, name=<optimized out>, data=0x7fffffffb440, datalen=<optimized out>, iType=<optimized out>)
    at /home/tkittel/work/mcstas-bug-aug19/nexus-4.4.3/src/napi5.c:2125
#9  0x00007ffff7fa8ae2 in nxiopendata_ (fid=0x55555557f8c0, name=0x55555556ec43 "Position") at /home/tkittel/work/mcstas-bug-aug19/nexus-4.4.3/src/napi.c:949
#10 0x000055555556b85d in _origin_24_setpos () at ./synts25.c:8918
#11 0x000055555556bbc5 in init () at ./synts25.c:9004
#12 0x000055555556d717 in mccode_main (argc=6, argv=0x7fffffffbb28) at ./synts25.c:10137
#13 0x000055555555d7f7 in main (argc=6, argv=0x7fffffffbb28) at ./synts25.c:6546
```

willend:
I think I may have just found it @tkittel - missing "closing" datasets after opening and adding data... 
(NXmakedata must be ended by NXclosedata)

Will run a full test and we shall see...

willend:
(So yes, seemingly improper use of the API - not the API itself.)

tkittel:
Following that trail, and I am indeed looking at that kind of code in the synts25.c :-)

tkittel:
This is just a quick glance, but look at the code below:

```
NXstatus NX5opengroup(NXhandle fid, CONSTCHAR * name, CONSTCHAR * nxclass)
{

  pNexusFile5 pFile;
  hid_t attr1, atype, iVID;
  herr_t iRet;
  char pBuffer[1024];
  char data[128];

  pFile = NXI5assert(fid);
  if (pFile->iCurrentG == 0) {
    strcpy(pBuffer, name);
  } else {
    sprintf(pBuffer, "%s/%s", pFile->name_tmp, name);
  }
```
So if I understand this correctly, if you do not close your groups, you are adding a too deep hierarchy of groups and the names will end up being slapped together ala `bla/origin_2/origin_3/origin_4" etc. And since 1024/25 ~= 40, you might indeed run out (especially if there are several items added for each of your 25 arms).

tkittel:
In another place (chasing with valgrind, now gdb) I find a potential problem if your "path" (inside the nexus file I assume) is deeper than 50. So if you have two layers per ARM, that would explain why it is around 25 you hit a snag.

tkittel:
```
  char pathStack[NXMAXSTACK][NX_MAXNAMELEN];
```
where `NXMAXSTACK` is 50.

tkittel:
In the relevant section of your synts25.c I see NXopendata but not NXclosedata, could that be it?

tkittel:
Only now I see that your previous comment indeed concerned opendata/closedata and not opengroup/closegroup like I had read it :-)

willend:
@tkittel yup, had the groups under control but close was missing in the datasets. Seems mostly functional now...

willend:
Closing, a (slightly) cleaner PR (https://github.com/McStasMcXtrace/McCode/pull/1687) supersedes this one...

PR #1667, opened 2024-08-05
Let mcviewtest become more of a user-tool:

willend:
* If no --testdir is provided, use PWD
* If no --reflabel is provided, use oldest subfolder in PWD
* Open result html by browser unless --nobrowse


willend:
Tests OK, merging.

PR #1668, opened 2024-08-05
Update McStas and McXtrace PDF manuals to indicate v. 3.5

willend:
Content unchanged, major doc revision coming post release...


willend:
Just updated PDFs, merging

PR #1672, opened 2024-08-09
tools/matlab: Octave use waitfor instead of pause.

farhi:
This small contrib is for the octave mcplot/mcdisplay. It replaces the use of `pause` by `waitfor` which allows to retain the interactive behaviour in plots. Closing the main window exits the plotter.


willend:
Tests ran OK + small and isolated to .m tools, OK to merge

PR #1674, opened 2024-08-16
Improvements to deployment of mcdisplay-webgl

willend:
* Ensure user ownership of js files (stored in Local ".mcstas/.mcxtrace" profile)
* Let mcdisplay-webgl kill node subprocess after a given timeout (defaults to 5 min)
* Use "junctions" on Windows to avoid copytree on the "node_modules" folder


willend:
Tests OK, merging

PR #1683, opened 2024-08-31
Allow to configure and use Monitor_nD buffersize from mcgui and mcrun.

willend:


willend:
All checks OK, merging

PR #1684, opened 2024-08-31
Leave copy of generated C code in dataset

willend:


willend:
All checks OK, merging

PR #1686, opened 2024-09-03
Feature adding to Monochromator_bent.comp

Lomholy:
This update of the instrument allows the user to set mosaicity to the crystal, as well as minor bug fixes to the calculations for the perfect crystal, making the results fit even better to the results from SIMRES.


willend:
(No rush, we can look together in a workshop break :) )

willend:
After discussion, the instr models need "perfect" `mosaicity=0` and we get output, as expected. :) 

@Lomholy I will add a commit later with improved instr files including `%Example` lines

PR #1690, opened 2024-09-11
New supermirror models from Hal Lee, ESS

willend:
Would benefit from a little more work on especially the comp header... We shall see if time is available to get that done before 3.5.0 release...


willend:
Should in no way interfere with the test, so simply merging...

PR #1695, opened 2024-09-13
E_smear -> v_smear in MCPL_input and related instruments

willend:
Implement fix for https://github.com/McStasMcXtrace/McCode/issues/1693


willend:
@g5t all tests passed, so I simply merge.

PR #1703, opened 2024-09-19
Add  -D_DARWIN_C_SOURCE on macOS

willend:
Apparently needed for functional `#include <stdio.h>` on macOS. 

Spotted when experimenting with / testing mccode-antlr on macOS, input from @g5t.


willend:
Tests OK http://tmp.mcstas.org/20240919_1727_35_output.html (for some reason there was initially an issue with Supermirror_thin_substrate, run in separate test for this reason)

PR #1704, opened 2024-09-20
Make use of mcpl_dump() in both McStas and McXtrace MCPL_output verbose mode

willend:
Sync McStas<->McXtrace MCPL_output.comp and use mcpl_dump() function to give a result summary


willend:
@g5t as per our chat.

g5t:
Other than potentially exposing the `mcpl_dump` settings, this _looks_ OK to me.
(I'm assuming that you copied from the McStas to McXtrace component correctly, and haven't checked it in detail)

willend:
I have run local tests myself via the Test_MCPL_output instruments. And this little change is certainly better than the previous hack, calling mcpltool via a system call.

PR #1708, opened 2024-09-24
[Fix] compilation error, if not behavior in Pol_guide_vmirror.comp

### LINKED ISSUE #1707: Pointer type mismatch in call to `Table_Read`
There are three calls to `Table_Read` in `Pol_guide_vmirror.comp` which pass a `MCNUM *` (aka `double *`) where a `char *` filename is expected. The first provides the component parameter `vector rpar {...}`
https://github.com/McStasMcXtrace/McCode/blob/66f71f44c85bbad0a94dd5efe7a18c98acd05b56/mcstas-comps/optics/Pol_guide_vmirror.comp#L180

This should generate a warning for all compilers, and generates an error on recent gcc.

The `Test_Pol_Guide_Vmirror.instr` example instrument uses the `Pol_guide_vmirror` component, and successfully compiles and runs in the nightly McStas tests. [Apparently generating warnings along the way](https://new-nightly.mcstas.org/todays-datafiles/mcstas-3.x-dev_NeXus_mpi_x_16_1e7_Linux/Test_Pol_Guide_Vmirror/compile_stdout.txt)

g5t:
The `Table_Read` code path in the `Pol_guide_vmirror.comp` component can not work, raises a compiler warning, and must not ever be used since the read tables are not accessed.

This PR half-fixes the problem by introducing new file name component parameters, thus removing the compiler warning.
If such a parameter was provided, however, the value(s) read from the table would still not be used.

By not producing a warning any longer, this fixes #1707

### DIFF
--- a/mcstas-comps/optics/Pol_guide_vmirror.comp
+++ b/mcstas-comps/optics/Pol_guide_vmirror.comp
@@ -46,6 +46,7 @@
 * 
 * rFunc: [1]       Cavity wall Reflection function
 * rPar: [1]        Cavity wall parameters array for rFunc, use with inputType = 0
+* rParFile: []     Cavity wall parameters filename for rFunc, use with inputType = 1
 * rR0: [1]         Cavity wall reflectivity below critical scattering vector, use with inputType = 2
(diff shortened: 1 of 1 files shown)


willend:
Thanks @g5t, merging

PR #1710, opened 2024-09-24
[Fix] missing pointer dereferences

### LINKED ISSUE #1709: Type mismatches in `Rotator` `Derotator` interaction
The `Rotator` defines a fixed-size `char` array as a component `DECLARE` parameter 
https://github.com/McStasMcXtrace/McCode/blob/66f71f44c85bbad0a94dd5efe7a18c98acd05b56/mcstas-comps/optics/Rotator.comp#L57
Then sets the name of a field that it adds to the `_particle` struct in its `INITIALIZE` block.

This is used by `Derotator`, through the `COMP_GETVAR3` macro to access the name of the field in the `_particle` struct, to access its value via `particle_getvar_void`

https://github.com/McStasMcXtrace/McCode/blob/66f71f44c85bbad0a94dd5efe7a18c98acd05b56/mcstas-comps/optics/Derotator.comp#L48-L51

The return value of `COMP_GETVAR3` is `char *[20]`, e.g., a pointer to the fixed-size array; yet it is assigned to a `char *` variable.

The return value of `particle_getvar_void` is always a `void *` and here it is cast as a `Rotation *`; but the first input to `rot_transpose` should be a `Rotation` instead.

In both cases de-referencing the pointer is necessary.

g5t:
Fixes #1709 compiler warnings; `Test_DiskChoppers2.instr` compiles and runs (output not verified)

### DIFF
--- a/mcstas-comps/optics/Derotator.comp
+++ b/mcstas-comps/optics/Derotator.comp
@@ -45,10 +45,11 @@ TRACE
   Rotation R;
 
   /* Name of Rotation uservar */
-  char* Rot_varptr = COMP_GETPAR3(Rotator, rotator, rot_var);
+  char* Rot_varptr = *COMP_GETPAR3(Rotator, rotator, rot_var);
 
   /* Invert rotation matrix for use in derotation */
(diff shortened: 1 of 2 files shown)


willend:
@g5t output verified, merging.

PR #1714, opened 2024-09-25
Fix contributed component type mismatches

### LINKED ISSUE #1713: Example instruments give type-mismatch compiler warnings
The example instruments `Test_Monochromator_bent.instr` and `Test_StatisticalChopper.instr` use components that cause compiler warnings under gcc due to passing a `char *` where a `const char *` is expected, or assigning an `int **` to an `int *` variable, respectively.

g5t:
Fixes #1713 by removing compiler warnings; correct behavior of test instruments not verified.


willend:
Meaningful patches @g5t, thanks.

I've run handheld tests with 3.5.1 and the patched components:
* The Mono instrument seems stable numerically
* The Statisticalchopper instrument performs differently, but in this case it would (to me) make sense to correct the "Expected output" later

i.e. merge - and I will postfix the test

PR #1716, opened 2024-09-26
Union restructure share includes

willend:
This PR reorganises the Union share codes to:
* Adhere to needs of the mccode-antlr code generator from @g5t which currently has issues with union-instrument
* - more specifically not `#include` inside another `#include`
* Combines Union_functions and Geometry_functions into union-lib
* Renames Union_initialisation to union-init
* Renames Union_last_functions to union-suffix


willend:
I ran a local test for instrument with "union" in the instr name
`mctest --testdir $PWD/mcstas-test --instr=union --nexus` (for some reason I chose NeXus output)

Output is available here:
http://tmp.mcstas.org/mcstas-union-test/

g5t:
Is this `@MCCODE_LIB@` magic still needed?
https://github.com/McStasMcXtrace/McCode/blob/3dd3b2de136054493b4a1a577fd976dc048e15ae/mcstas-comps/union/Union_master.comp#L63

I guess that `MCCODE_LIB/share` is `mcstas-comps/share` in the repository, which gets installed (and searched) along side the components and example instrument files, right?

willend:
My feeling is that it may still be needed. 

Files from the mentioned directories mostly use `%include` brutal concatenation, whereas Mads uses `#include`, i.e. a precompiler directive. I am not sure the those directories otherwise make it to `-I` during compilation.

But feel free to test + adjust if it works without.

g5t:
Only `NCrystal_process.comp` in `mcstas-comps/union` uses any `#include` preprocessor directives -- and the two files included that way are `union-lib.c` and `union-init.c`, both of which are `%include`d in the `Union_master` components. So they should always be present for the process component, right?

willend:
Yes I tend to think you are right about that. If you have time, maybe try a one-off test with the relevant instr to confirm?

g5t:
The `#include` statements in `NCrystal_process.comp` were guarded by an `#ifndef Union` preprocessor directive.
I've replaced them by `%include` code-generator directives (with the same guard), which works, but I think is ultimately not the full solution.

Instead, assuming _all_ Union components are required to come after the `Union_init.comp`; all other Union components should replace their
```C
#ifndef Union
#define Union ...
%include union-lib.c
%include union-init.c
#endif
```
with
```C
#ifndef Union
#error "A Union_init component is required before XXX"
#endif
```
to enforce the relationship at compilation time. 
Currently, the only check is done in `Union_init.comp` which produces an error message and exits _at runtime_ only.

g5t:
A major benefit to removing the `%include union-lib.c` and `%include union-init.c` statements from each Union component is improved compilation time.

Since those are _code-generator_ directives, they place the full contents of both files inside of `#ifndef Union` guards for every Union component used in a file. (Preprocessor directives would typically be matched with their own `#ifndef` or `#pragma once` guards)
This significantly increases the size of the file, which increases preprocessor parsing time.

Since the `#ifndef Union` must _always_ be false for a working instrument, those `%include` statements never led to useful code inclusion.

All Union components now have `#error` directives, but compilation is likely to fail _anyway_ without a `Union_init` component due to the reliance on `COMP_GETPAR3(Union_init, ...`

Compiling the two `Union_NCrystal*` example instruments on my machine went from 97 seconds with the `%include` directives to 32 seconds with the `#error` directives.

mcstas-comps/share/union-init.c
g5t, line 1: This file is actually redundant. The 'global' variables reside inside of the `Union_init` component parameters `struct`. This was `%include`d previously but never used due to `#ifndef Union` guards and the requirement for a `Union_init` component instance to be defined in a working instrument.

Unless if a push toward `Union_init`-less functionality is envisioned, this file should be removed.

g5t (approved):
The current version of Union components works* with `mccode-antlr` (* compiles and runs, but correctness has not been verified).

Given the changes I've made, maybe @willend should review the current state of things and verify that everything still works in a more-official manner?

willend:
@g5t I think this all looks good and "should work". I will merge for now and accept to take care of potential collateral damage  later.

mads-bertelsen:
Sorry I am late to the party, was out for a bit. Great you found a way to reduce there code size given to the precompiler and reduced the compile time!

I don't quite understand why the two geometry and functions files were merged to one, thought it was nice to have that context. They had both grown to a size where one needs to search for the desired function, so it doesn't matter much i suppose.

PR #1726, opened 2024-10-08
Fix of Single_crystal reuse optimization

mads-bertelsen:
Fixed issue in Single_crystal where neutrons with similar initial velocity vector would reuse wrong data from previous neutron. Allocated a seperate data structure for the initial search for reflections which avoids this issue. Fixed a test to be sensitive to such an issue in the future.

Suggestions on how to make this GPU compatible have been made, so this PR should not be merged yet.


willend:
@mads-bertelsen thanks. Actual feedback to this might not arrive until ~ November since busy with other stuff. :-)

willend:
Merging this solution for now - deferring GPU uservars solution until later, as also mentioned in https://github.com/McStasMcXtrace/McCode/issues/1725

PR #1727, opened 2024-10-09
PowderN: fixing tth_sign argument

farhi:
The `tth_sign` argument seems to not have been effective. I also added it to the McXtrace comp.


farhi:
I think all is now fine...

PR #1738, opened 2024-10-11
McStas: Fix #1732: Single_crystal powder mode was broken by #1521

### LINKED ISSUE #1732: McStas: single_crystal(powder=1) is not functional. McXtrace is OK
The `single_crystal(powder=1)` component does not provide the proper results. The equivalent McXtrace implementation works nicely. 
I recommend comparing the two components, and fix the neutron side...

farhi:
All is in the title.
The bug seen for #1521 is not really a bug. 
There are two propagation steps in the process:

1. propagate to the sample surface
2. rotate the crystal (powder mode)
3. scatter
4. propagate in the "rotated" crystallite (as it takes place in reality)
5. put back the crystal in its initial state

What was seen in the issue #1521 is that the final propagation assume there is enough matter in the crystallite to actually still propagate in the sample volume. In some cases indeed, this propagation can get the neutron outside the sample volume after de-rotating.

A better fix to #1521 would perhaps be to check after exiting the crystal, and potentially propagate back onto the surface. using the final velocity and position.
For the time being, I consider it is better to get a few neutrons outside the volume, than not having a powder mode ;-)

### DIFF
--- a/mcstas-comps/examples/Tests_optics/Test_Monochromators/Test_Monochromators.instr
+++ b/mcstas-comps/examples/Tests_optics/Test_Monochromators/Test_Monochromators.instr
@@ -24,6 +24,7 @@
 * %Example: Mono=5 Detector: psd1_I=8.9e-05
 * %Example: Mono=6 Detector: psd1_I=1.0e-04
 * %Example: Mono=6 PG=0.15 Detector: psd1_I=2.8e-05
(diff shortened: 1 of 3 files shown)


willend:
Hi @farhi,

I ran a few hand-held tests with the comp and instr from the PR and some other key SX-oriented instruments. - Looks OK
https://new-nightly.mcstas.org/mcstas-SX-test/mcstas-test_output.html

Merge

PR #1745, opened 2024-10-16
Fix for https://github.com/McStasMcXtrace/McCode/issues/1741

willend:
This in effect means:

* Partial revert of https://github.com/McStasMcXtrace/McCode/commit/f9c844207a140d2e12aed13922c216b4a6a94c36 and https://github.com/McStasMcXtrace/McCode/commit/bc505762c2ee0beda817c74f03eb4b521ec28745
* Documenting that `tth_sign` only works **together** with `d_phi` 

I further found that weigh corrections were not correctly applied with `tth_sign`, powder lines became a factor of 2 too strong.


farhi:
Thanks Peter, Would be nice to port as well to X-rays ???

willend:
Yup, I will have a look at that also.

willend:
(But may not happen until next week, so will simply keep the issue open)

PR #1755, opened 2024-11-04
Work toward fixing SX issue from @minghuisvn

willend:
As mentioned towards the end of https://github.com/McStasMcXtrace/McCode/issues/1728 giving direct / reciprocal vector-input to McXtrace (and McStas) gave weird results.

This set of commits suppresses file-provided lengths/angles if vectors are present.


willend:
Checks OK - merge

PR #1760, opened 2024-11-14
Fix for "interesting" SPLIT + GROUP issue.

willend:
Found by Torben R. Nielsen: An instrument with SPLIT and a subsequent GROUP of monitors would "lose intensity" and events. After a long analysis with Erik we found that this was due to use of ABSORB; in the final GROUP member (but in the raytrace function) yet still meaning exit().

Hence, non-scattered neutrons (or e.g. n'th pass of the SPLIT) would always exit not only the group, but the full raytrace() fct.


willend:
All checks OK, merging.

PR #1762, opened 2024-11-18
Introduction of order=0.5 mode in Single_crystal

mads-bertelsen:
The optimization in Single_crystal to reuse calculations for the previous ray when they are identical was supposed to speed up the calculation with a large factor, but in practice it was found to only reach a speed up of about 2 even with large split numbers. This was explored as a tangent in this issue: https://github.com/McStasMcXtrace/McCode/issues/1725

The reason is that the expensive hkl_search is performed both for the ray entering the crystal (which can be reused as the wave vector will be exactly the same as the previous ray when using split), and for when leaving the crystal, hence at most half of the work can be reused.

Here a new mode is introduced, order=0.5, where the attenuation from coherent scattering as the ray leaves the crystal is simply ignored, meaning the number of hkl_search calls can be reduced with a factor equal to the split number.

Here are results for lucine with different values for the order:
lucine_order_comparison_wide
lucine_order_comparison_narrow

As well as rubredoxin: 
output_narrow_rubredoxin

The new mode, order=0.5 is stable when changing split value as expected:
order_half_split_stable

The execution time as a function of order and split is shown in the graf below:
execution_time_orders
This is for a target ncount after the crystal, so split 100 would be done with a factor 100 less rays than split 1. The fastest run with order=0.5 is barely more than a second, while the slowest runs with higher order is about 100 seconds.

The speedup between the different order settings can be seen below, the limit of 2 for order=1 is clearly seen.
speedup_orders

Here a speedup of almost a factor of 30 is seen. The increase in speed is suspected to grow for higher split numbers, but will saturate as the time spent elsewhere in the simulation starts to become the bottleneck.

### DIFF
--- a/mcstas-comps/samples/Single_crystal.comp
+++ b/mcstas-comps/samples/Single_crystal.comp
@@ -19,6 +19,7 @@
 * Modified by: PW, June 2017: Doc updates
 * Modified by: PW, Feb 2018:  GPU edits
 * Modified by: EF, June 2023: can now read CIF files
+* Modified by: MB, November 2024: Order 0.5 optimization with large split numbers
 *
 * Mosaic single crystal with multiple scattering vectors, optimised for speed
 * with large crystals and many reflections.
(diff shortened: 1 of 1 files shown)


willend:
@mads-bertelsen thanks for this, most welcome. I will have a look at adjusting / adding to tests including Single_crystal later.

PR #1764, opened 2024-11-19
Fix of order system in Single_crystal

mads-bertelsen:
It was realized that the order system in Single_crystal in general performed unnecessary calculations, mainly calculating the coherent cross section, and then not using it when the ray leaves the crystal. This was fixed, and this made recent introduction of a order=0.5 mode and double bookkeeping of the heavy Tau_list unnecessary, so these were removed.

This update retains the speedup of order=0.5 for order=1 and reduces the memory footprint significantly for large systems.
Figure 3
The above graph was done with a relatively small ncount / problem size, the ceiling for speed up is increased with these parameters, speed ups of factor 100 are still observed with sufficiently large problems.

There is a significant difference between order=0 (all multiple scattering) and order=1, there are still improvements that could be made without sacrificing performance, for example allowing incoherent scattering to higher orders even when the coherent part is disabled.
Figure 8

Some changes were made in the propagation for Powder and PG mode that can lead to the ray leaving the sample during normal propagation. This is reversed to a more secure version. To show this works as intended, there is a test instrument that compares PowderN and Single_crystal with the powder mode. This fails as they are not exactly alike, meaning there is more debugging to do.

### DIFF
--- a/mcstas-comps/examples/Tests_samples/Test_Powders/Test_Powders.instr
+++ b/mcstas-comps/examples/Tests_samples/Test_Powders/Test_Powders.instr
@@ -1,0 +1,100 @@
+/*******************************************************************************
+*         McStas instrument definition URL=http://www.mcstas.org
+*
+* Instrument: Test PowderN output
+*
+* %Identification
+* Written by: M. Bertelsen
+* Date: November 19 2024
+* Origin: ESS DMSC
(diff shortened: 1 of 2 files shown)


willend:
@mads-bertelsen thanks all of this follow-up work. I think a comparison across to McXtrace Single_crystal would also be very nice.

PR #1767, opened 2024-11-20
Added order_extra feature for Single_crystal

mads-bertelsen:
Added extra_order feature to Single_crystal, allowing extra orders of multiple scattering that ignores coherent scattering which is slow for large unit cells.

In a case as below, order=0 does not give satisfactory approximation.

order_approach_ref

Using higher orders provides less speedup as a function of split as show below.

Figure 3

This update introduces the option to allow a limited number of additional multiple scattering that ignores the coherent part of the cross section which is the computationally expensive part of the algorithm. 

Figure 199-2
Figure 199-3

Using this feature with order=1 quickly provides a very good approximation with less computation time. Below we see speed ups compared to order=0, all with SPLIT 100. Using extra_order provides a better approximation, but of course does take computation time, though much faster than increasing the normal order parameter.

Figure 204

It can be a bit complex to choose the right combination of SPLIT, order and extra_order. The plot below shows many runs with speed up relative to a reference simulation and a chi square comparison to the reference run (here extra_order is called order_inc / "oi" in label, order is "o",  SPLIT as "s"). For this system, leucine, using order=1, high split and extra_order somewhere between 3 and 6 would provide a good balance between accuracy and computation time.

Figure 191

Will update this pull request with equivalent feature for McXtrace.

### DIFF
--- a/mcstas-comps/samples/Single_crystal.comp
+++ b/mcstas-comps/samples/Single_crystal.comp
@@ -182,6 +182,7 @@
 * cz: []                                                  c on z axis
 * reflections: [string]                                   File name containing structure factors of reflections (LAZ LAU CIF, FullProf, ShelX). Use empty ("") or NULL for incoherent scattering only
 * order: [1]                                              Limit multiple scattering up to given order (0: all, 1: first, 2: second, ...)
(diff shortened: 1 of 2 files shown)


mads-bertelsen:
McXtrace version added.

PR #1775, opened 2024-11-23
Reorder config dialogue for "small screens"

willend:
Edits to make dialogue double width and half height.


farhi:
An other solution would be to separate the plotters, which are often changed, from other less used preferences. In this case we would have two dialogues. Or add a slider on the right side ?

Issue #1675, opened 2024-08-19
McXtrace: cif2hkl does not work


minghuisvn:
I installed the McXtrace 3.2 metapackage in MS windows 11, following the instruction as follows:
"https://github.com/McStasMcXtrace/McCode/tree/main/INSTALL-McXtrace/Windows". When i run "cif2hkl --help" command in mcxtrace-shell-3.2, a "could not found libgfortran-5.dll" error occurred. Then i installed the strawberry perl software in windows. Then i rerun the same command "cif2hkl --help" , the prompt windows outputted nothing. What should i do to solve this problem?
![image](https://github.com/user-attachments/assets/795557fa-d3be-4c34-9601-6462534e0c96).  
  I used the cif2hkl of fullprof to convert cif files to hkl files before, but there were always some problemes when simulating single crytal diffraction using hkl files generated through cif2hkl tool of fullprof suit. So i decide to use cif2hkl tool within McXtrace software.
   I tried another method, i.e., installing cif2hkl in the WSL of windows through make compilation tool. After the installation, "cif2hkl --help" command is OK. I tried to convert Al.cif to Al.lau by command "cif2hkl --mode XRA --xtal --lambda 1.0  -o Al.lau  Al.cif", but unexpectedly, the column F2 is always zeros in the obtained Al.lau files. The cif and resulted lau files are uploaded to the attachment.
![image](https://github.com/user-attachments/assets/553f4d26-9fe5-4d1f-90cb-3a54dcff852f)
[Al.cif.txt](https://github.com/user-attachments/files/16668187/Al.cif.txt)
[Al.lau.txt](https://github.com/user-attachments/files/16668188/Al.lau.txt)

farhi:
Hello,
You are right, the XRA generation often provides F2=0, which is not the case for NEU mode. 

I've been discussing with the CrysFML authors, but they did not seem to be willing to correct the bug, see:
- https://code.ill.fr/scientific-software/crysfml/-/issues/6

I'll see if I can fix this myself...

farhi:
Just found the error.

farhi:
The new fixed version is available at:
- https://gitlab.com/soleil-data-treatment/soleil-software-projects/cif2hkl

where you can find a deb package to install e.g. in WSL.
- https://gitlab.com/-/project/21820312/uploads/2a62300f84cb044900bca88ccdaed2b1/cif2hkl_1.4.5_amd64.deb

PR #1777, opened 2024-11-24
PowderN: fix warning message repeated for ever

farhi:
The flag_warning was not properly incremented and summed-up in the line_info struct.
Then every particle would always start with a 0 warning counter, creating very long output, that hinder computation efficiency.
Now fixed.


farhi:
A new major fix for cif2hkl, even if not included anymore.
Close #1675

PR #1786, opened 2024-12-03
Print warning when overwriting COPY EXTEND %{%} block from instance

willend:
Candidate solution to https://github.com/McStasMcXtrace/McCode/issues/1785

The attached instrument [`COPY_EXTEND_1`](https://github.com/user-attachments/files/17997589/COPY_EXTEND_1.instr.txt) has a `COMPONENT comp2=COPY(comp1)` instance with an `EXTEND %{%}` block, which thus overwrites the `EXTEND %{%}` that arrives via the `COPY(comp1)` clause:

```
mcstas COPY_EXTEND_1.instr 

-----------------------------------------------------------
WARNING (COMPONENT comp2): Existing (COPY) EXTEND block:

  EXTEND %{
    // An extend block
    dummy=_comp->index;
  %}

will be been overwritten by:

  EXTEND %{
    // An other extend block
    dummy=2*_comp->index;
  %}
-----------------------------------------------------------

-----------------------------------------------------------

Generating single GPU kernel or single CPU section layout: 

-----------------------------------------------------------

Generating GPU/CPU -DFUNNEL layout:

-----------------------------------------------------------
CFLAGS=
```

I further attach [`COPY_EXTEND_2`](https://github.com/user-attachments/files/17997534/COPY_EXTEND_2.instr.txt) which has a `COMPONENT comp2=COPY(comp1)` with an `EXTEND %{%}` block,  but where `COPY(comp1)` includes no `EXTEND` block:

```
mcstas COPY_EXTEND_2.instr 

-----------------------------------------------------------

Generating single GPU kernel or single CPU section layout: 

-----------------------------------------------------------

Generating GPU/CPU -DFUNNEL layout:

-----------------------------------------------------------
CFLAGS=

```

### DIFF
--- a/mcstas/src/instrument.y
+++ b/mcstas/src/instrument.y
@@ -1279,14 +1279,14 @@ component: removable cpuonly split "COMPONENT" instname '=' instref
 	      fprintf(stderr, "  %s",line);
 	    }
 	    list_iterate_end(liter);
-	    fprintf(stderr, "\n  %%}\n");
+	    fprintf(stderr, "  %%}\n");
 	    fprintf(stderr, "\nwill be been overwritten by:\n");
 	    fprintf(stderr, "\n  EXTEND %%{\n");
 	    while((line2 = list_next(liter2))) {
 	      fprintf(stderr, "  %s",line2);
 	    }
 	    list_iterate_end(liter2);
-	    fprintf(stderr, "\n  %%}\n-----------------------------------------------------------\n");
+	    fprintf(stderr, "  %%}\n-----------------------------------------------------------\n");
 	  }
 	  comp->extend= $13;  /* EXTEND block*/
 	}
--- a/mcstas/src/instrument.y
+++ b/mcstas/src/instrument.y
@@ -1270,18 +1270,18 @@ component: removable cpuonly split "COMPONENT" instname '=' instref
         if ($13->linenum) {
 	  if (comp->extend->linenum>0) {
 	    fprintf(stderr, "\n-----------------------------------------------------------\n");
-	    fprintf(stderr, "WARNING (COMPONENT %s): Existing (COPY) EXTEND block:\n", comp->name);
+	    fprintf(stderr, "WARNING: Existing (COPY) EXTEND block in COMPONENT %s:\n", comp->name);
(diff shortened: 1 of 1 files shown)


mcstas/src/instrument.y
```diff
@@ -1267,7 +1267,29 @@ component: removable cpuonly split "COMPONENT" instname '=' instref
...
+	    List_handle liter = list_iterate(comp->extend->lines);
+	    List_handle liter2 = list_iterate($13->lines);
+	    char *line, *line2;
+	    fprintf(stderr, "\n  EXTEND %%{\n");
+	    while((line = list_next(liter))) {
+	      fprintf(stderr, "  %s",line);
+	    }
+	    list_iterate_end(liter);
+	    fprintf(stderr, "  %%}\n");
+	    fprintf(stderr, "\nwill be been overwritten by:\n");
```
g5t, line 1283: I think a slightly different wording of the warning might make it clearer. Something like
```
	    fprintf(stderr, "WARNING: Existing (COPY) EXTEND block in COMPONENT %s:\n", comp->name);
	    List_handle liter = list_iterate(comp->extend->lines);
	    List_handle liter2 = list_iterate($13->lines);
	    char *line, *line2;
	    fprintf(stderr, "  EXTEND %%{\n");
	    while((line = list_next(liter))) {
	      fprintf(stderr, "  %s",line);
	    }
	    list_iterate_end(liter);
	    fprintf(stderr, "  %%}\n");
	    fprintf(stderr, "\nis overwritten by:\n");
```
(Plus I would either remove the extra line breaks, like I've done here, or remove the two-space indentation of the printed `EXTEND` blocks)
e.g.,
```cmd
WARNING: Existing (COPY) EXTEND block in COMPONENT named_instance:
  EXTEND %{
  ...
  %}
is overwritten by
  EXTEND %{
  ...
  %}
```

g5t (approved):
This change accomplishes the goal to warn users when they overwrite an inherited component instance `EXTEND` block set via `COPY`.
I think the message could be clearer, but my opinion is not necessarily correct.

mcstas/src/instrument.y
willend: Thanks @g5t, will implement that.

Issue #1817, opened 2025-01-22
Variable length array allocation in `polyhedron.c` is not supported on all platforms


g5t:
Allocating arrays with sizes determined by the value of a struct member, as in,
https://github.com/McStasMcXtrace/McCode/blob/bdd87f6249fcae1f8c2bb53a056443f12772b86b/mcstas-comps/share/polyhedron.c#L661-L666
fails under Microsoft's C/C++ Optimizing Compiler Version 19.42.34435 for x64 

```
E  cl.exe /DWIN32 /D_WINDOWS /Zi /O2 /Ob1 /DNDEBUG g\source.c /W0 /link /out:"g\source.exe"
E  failed with output
E  source.c
E  g\source.c(8491): error C2057: expected constant expression
E  g\source.c(8491): error C2466: cannot allocate an array of constant size 0
E  g\source.c(8491): error C2133: 'dtime': unknown size
E  g\source.c(8492): error C2057: expected constant expression
E  g\source.c(8492): error C2466: cannot allocate an array of constant size 0
E  g\source.c(8492): error C2133: 'dpoint': unknown size
E  g\source.c(8493): error C2057: expected constant expression
E  g\source.c(8493): error C2466: cannot allocate an array of constant size 0
E  g\source.c(8493): error C2133: 'time': unknown size
E  g\source.c(8494): error C2057: expected constant expression
E  g\source.c(8494): error C2466: cannot allocate an array of constant size 0
E  g\source.c(8494): error C2133: 'point': unknown size
E  g\source.c(8495): error C2057: expected constant expression
E  g\source.c(8495): error C2466: cannot allocate an array of constant size 0
E  g\source.c(8495): error C2133: 'plane_number': unknown size
E  g\source.c(8496): error C2057: expected constant expression
E  g\source.c(8496): error C2466: cannot allocate an array of constant size 0
E  g\source.c(8496): error C2133: 'type': unknown size
```

This method of variable length array allocation would appear to be part of the C99 standard, but [according to Microsoft](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version)
>The compiler doesn't implement several required features of C99, so it isn't possible to specify C99 conformance, either.

Trying either of the supported language specifications, `/std:c11` or `/std:c17` does not allow an instrument using the polyhedron intersection code to compile for this tested version of the Microsoft C compiler.

Replacing this construction with paired allocation and freeing memory should be more cross-platform friendly.

[...]

PR #1819, opened 2025-01-22
Remove variable length array allocations for #1817

g5t:
Replace variable length array allocation with initializer-list allocation.

This is a partial fix for #1817 but does not correct all instances of variable length array allocation.


willend:
@g5t thanks for the PR, edits very reasonable.0 Merging on the basis that all test are now flagged green.

PR #1821, opened 2025-01-23
Remove NCrystal_sample.comp possible out-of-bounds memory access

g5t:
`memset(&params,0,sizeof(geoparams))` would cause an out-of-bounds memory error if `sizeof(geoparams)` > `sizeof(params)`; and anyway does not appear to do what is intended.

### DIFF
--- a/mcstas-comps/contrib/NCrystal_sample.comp
+++ b/mcstas-comps/contrib/NCrystal_sample.comp
@@ -204,7 +204,7 @@ INITIALIZE
 
   //for our sanity, zero-initialise all instance-specific data:
   memset(&params,0,sizeof(params));
-  memset(&params,0,sizeof(geoparams));
+  memset(&geoparams,0,sizeof(geoparams));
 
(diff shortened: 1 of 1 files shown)


willend:
@g5t thanks. Could you have a look if the related `union/NCrystal_process.comp` may have similar potential issues? 

I will consider this PR a work in progress until you 'say when' - OK?

g5t:
`NCrystal_process` only zeros the `params` structure (not surprising, since the geometry is handled elsewhere in Union) 
https://github.com/McStasMcXtrace/McCode/blob/d7e4564a06aefa7c53e4edeb00ded26d80a2322a/mcstas-comps/union/NCrystal_process.comp#L289

And, as far as I could find, `memset` is not used in any other components. So this could be merged safely.

(Alternatively, dropping the second `memset` is probably fine since a) it wasn't being zeroed before anyway and b) the call to `ncrystalsample_initgeom` on the next line will either initialize all fields of the struct _or_ cause an error due to mismatched component parameters.)

willend:
Interestingly the patch still doesn't seem to fix the intermittent issue/failed test with MPI on macOS. Will merge anyhow.

PR #1828, opened 2025-01-29
NeXus: sort comp instances by index and include comp parameter information

willend:


willend:
"Enumerated" component labels for sorting like in the instrument:
Screenshot 2025-01-30 at 10 51 36

Component type included:
Screenshot 2025-01-30 at 19 03 56
Screenshot 2025-01-30 at 19 06 56

Parameter list included pr. component, with type, default and "value as in instrument file". 
Screenshot 2025-01-30 at 19 04 20

Issue #1833, opened 2025-01-31
McDisplay pipetools trace output parsing is too restrictive


g5t:
The regular expression used to identify the instrument information lines are too restrictive.
https://github.com/McStasMcXtrace/McCode/blob/da4bbb289dd656d3b2f4952f99c16d17556d8cd7/tools/Python/mccodelib/pipetools.py#L104

Valid paths and instrument file names are ignored, leading to an error in `InstrObjectConstructor`, presumably because the rule at
https://github.com/McStasMcXtrace/McCode/blob/da4bbb289dd656d3b2f4952f99c16d17556d8cd7/tools/Python/mccodelib/instrparser.py#L135
is never matched due to the second output line not matching above.

I would expect all test cases here to pass the regular expression match: https://regex101.com/r/JJmapQ/2
but, notably, paths with a `.` or filenames without the extension `.instr` are ignored.

Changing the regiular expression in `pipetools.py` necessitates the same change to the parsing rule for an `ABSPATH` keyword in `instrparser.py`:
https://github.com/McStasMcXtrace/McCode/blob/da4bbb289dd656d3b2f4952f99c16d17556d8cd7/tools/Python/mccodelib/instrparser.py#L91-L93

## Suggested solution
Update the `pipetools.py` regular expression to [`Instrument \'[\w\-\.]+\' \([/\w\\\:\-\.]+(\.instr)?\)`](https://regex101.com/r/wrhfgU/1)

Plus update the `INSTRNAME` and `ABSPATH` token rules in `instrparser.py` to  `r'\w[\w\-0-9\.]*'` and `r'[/\w\\\:\-\.]+(\.instr)?'`, respectively.

PR #1834, opened 2025-01-31
[Fix] too-restrictive McDisplay instrument line parsing

g5t:
My preferred solution to issue #1833


willend:
Manual build and and mcdisplay test with both mcstas-antlr and mcstas from main checks out. Merging.

PR #1847, opened 2025-02-09
Gzip biggest McXtrace datafiles

willend:
Replace big ascii datafiles (without ref in comp/instr) with gzip'ed versions

Awaiting comments from Erik before merge.

### DIFF
--- a/meta-pkgs/windows/environment.yml
+++ b/meta-pkgs/windows/environment.yml
@@ -8,6 +8,7 @@ dependencies:
   - m2-msys2-runtime
   - m2-which
   - m2-sed
+  - m2-gzip
   - m2w64-gcc
   - m2w64-gsl
   - msmpi


ebknudsen:
There is no direct reference to the bilayer coatings in the relevant instrument files (e.g. ATHENA_cfgA_1mm.instr). The exact coating data file is supplied as a string command line argument. I see no sane reason why compressing these files would be a problem. In fact a single example file is enough. The datafiles themselves were originally by an IMD program.

Suggestions:
1. make read_table-lib gzip-transparent, like mcpl.
2. make sure the ref-lib interface does not trip on extra header lines describing what the data file contains (not sure if this is the case at present). For instance the IMD-program that generated them, could be useful for someone trying to reproduce something.

willend:
Impatient me... I've simply added them gzipped for now. :-) I might take your comments above @ebknudsen to a new issue to fix "properly" in some distant future ;-)

PR #1852, opened 2025-02-12
meta-pkgs:deb:mcxtrace: add libgsl-dev as dependency

farhi:
Add dependency on libgsl for debian packages.


farhi:
The failing test has nothing to do with the PR.

PR #1874, opened 2025-02-26
NCrystal default branch was renamed from master to main

tkittel:
Although the cmake fetch ncrystal is a bit obsolete, at least it should refer to the correct branch name :-)


willend:
OK, we start there

Issue #1866, opened 2025-02-20
Periodic segfaults NCrystal+macOS+openmpi5


willend:
See e.g. https://github.com/mccode-dev/McCode/actions/runs/13442024853

```bash
+ NUM_MPI=1
+ ../install_mcstas/bin/mcrun --mpi=1 NCrystal_example.instr 'sample_cfg=Y2O3_sg206_Yttrium_Oxide.ncmat\;density=0.6x'
INFO: No output directory specified (--dir)
INFO: Using directory: "NCrystal_example_20250220_182358"
INFO: Regenerating c-file: NCrystal_example.c
CFLAGS= @NCRYSTALFLAGS@ -DFUNNEL 

INFO: Recompiling: ./NCrystal_example.out
-----------------------------------------------------------

Generating single GPU kernel or single CPU section layout: 
-> SPLIT N at component monochromator
-> SPLIT N at component powder_sample

-----------------------------------------------------------

Generating GPU/CPU -DFUNNEL layout:
Component monochromator is NOACC, CPUONLY=1
-> FUNNEL mode enabled, SPLIT within buffer.
-> SPLIT within buffer at component monochromator
-> GPU kernel from component mono_out
Component powder_sample is NOACC, CPUONLY=1
-> FUNNEL mode enabled, SPLIT within buffer.
-> SPLIT within buffer at component powder_sample
-> GPU kernel from component powder_pattern_detc

-----------------------------------------------------------
./NCrystal_example.c:3772:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
} /* siminfo_init */
^
1 warning generated.
ld: warning: ignoring duplicate libraries: '-lmpi'
ld: warning: search path '/lib' not found
INFO: ===
[NCrystal_example] Initialize
NCrystal: McStas sample component(s) are using version 4.0.2 of the NCrystal library.
*** TRACE end *** 

# Last I/O Error: Operation timed out
Save [NCrystal_example]
--------------------------------------------------------------------------
Detector: powder_pattern_detc_I=4.37203e-11 powder_pattern_detc_ERR=1.5805e-13 powder_pattern_detc_N=76565 "powder_pattern_detc.dat"
prterun noticed that process rank 0 with PID 9630 on node Mac-1740074734169 exited on

signal 3 (Quit: 3).
Finally [NCrystal_example: NCrystal_example_20250220_182358]. Time: 8 [s] 
--------------------------------------------------------------------------

[...]

PR #1884, opened 2025-03-03
Fix issue 1883 by setting mpirun --mca params

willend:
Also ported to basictest and McStas tests


willend:
One hung process, seems to relate to https://github.com/mccode-dev/McCode/issues/1866. Leaving as is and merging to have a bit more output data going forward.

PR #1886, opened 2025-03-06
Full adaptation of "spin-component nullification" after "spin up/down measurement"

willend:
Logic and procedure from Pol_mirror.comp ported to other polarised components.


willend:
Relevant instruments (manually) tested for reasonable output on Linux CPU and GPU:
Screenshot 2025-03-06 at 11 33 21

PR #1890, opened 2025-03-07
MacInstall: add comment on forcing open permission with command

farhi:
I added a command to do the same as with the multiple dialogues, in a single shot.


willend:
Good point, even though I think this may be beyond some mac users. :-) I've copied across the remark to the McStas side also.

PR #1919, opened 2025-04-03
mcxtrace:powderN: Update PowderN.comp: reduce nb of printed warnings

farhi:
print 100 -> 10 warnings


willend:
Added similar reduction in warnings for the McStas side (McXtrace Single_crystal already had 10 max)

willend:
Merging. @farhi I will also update the hackathon branch

Issue #1923, opened 2025-04-06
`printf` compilation error in `MCPL_input`


g5t:
A recent(ish) change to `MCPL_input.comp` introduced a call to `printf` which mistakenly uses the syntax of `fprintf`, that is 
https://github.com/mccode-dev/McCode/blob/ceecb027d37e21b7356c046c07e564c40acfec79/mcstas-comps/misc/MCPL_input.comp#L140
is used, causing a compilation error on at least one system, where either of `fprintf(stdout, "%s", tmpstr);` or `printf("%s", tmpstr);` would have worked.

willend:
@g5t was fixed and merged. closing.

PR #1924, opened 2025-04-06
Update MCPL_input.comp

g5t:
One of the two possible fixes for #1923 -- `fprintf` chosen in case `stdout` can somehow be replaced by a normal file somewhere else, and this output would be expected there.


willend:
@g5t thanks for spotting that one. I am puzzled that neither CI tests nor nightlies failed on this?

PR #1939, opened 2025-04-11
DONT MERGE YET: PowderN: remove a static variable declare from SHARE section

farhi:
The **PowderN** components have a misplaced and redundant `off_struct offdata` declaration that has nothing to do here. It is actually also in the DECLARE block, so that all is fine.

PR #1942, opened 2025-04-12
McXtrace: Fluorescence handling ala PowderN

willend:
Smaller chunk of https://github.com/mccode-dev/McCode/pull/1939 from @farhi


willend:
@farhi merged

PR #1944, opened 2025-04-12
mcrun: remove gcc command as it is displayed with --verbose

willend:
Smaller chunk of https://github.com/mccode-dev/McCode/pull/1939


willend:
@farhi merged

PR #1951, opened 2025-04-19
McXtrace: samples: add Fluo+PowderN component

farhi:
This is a kind of merge of PowderN and Fluo/Abs processes.
A test instr is also provided.


farhi:
PS : I made a few wrong edits, which were reverted. 
It seems we have an OpenCL version of hkl_search for mcstas/SX, which could be nice to look at for further performance boost in SX.

common/lib/share/interoff-lib.c
willend, line 1: Just versioning defines, OK

common/lib/share/read_table-lib.c
willend, line 1: Simple defines, OK

willend:
Hi @farhi ok. I will revert most of the edits (if not all) to PowderN:
* "const char" filename was introduced when support was added for `msvc` / `conda`
* The local trace `flag_warning` allows to run seamlessly on GPU, atomic addition at the end of `TRACE`

mcxtrace-comps/samples/PowderN.comp
willend, line 1: Added commit to bring back local TRACE `flag_warning` and let filename be `const char` in `cif2hkl` fct. courtesy of `msvc` on Windows.

willend (approved):
Playing with the review / approval just for fun. :)

willend:
@farhi will let the test run through, then merge

farhi:
Hi Peter !

Regarding the PowderN warning (did not hit from inside), I have the feeling that the `flag_warning++` goes through, but as they are followed by ABSORB, the actually never increment the global `line_info.flag_warning`. 
The result is that all warnings are displayed...

PR #1952, opened 2025-04-20
PowderN and SX fixes: cif2hkl now searches correctly files using Open_File/real path, and sets 'barns=1'

farhi:
A substantial `cif2hkl` fix:
- properly get full path of CIF file when e.g. from the 'data', via `Open_File(-,-,char *realpath)`  (required to provide full path before)
- set `barns=1`

I also fixed (again ?) the long standing many 'particle did not hit from inside' warnings.
This is because the global flag variables are not properly updated after the TRACE section. They remain at 0 even when local flag variables are incremented. Perhaps there something wrong, if so, tell me ;-)


farhi:
OK, I reverted the `flag_warning` stuff in PowderN and SX, as well as the `const char` in `cif2hkl`.
Here are more commits regarding Fluorescence+PowderN samples for McXtrace.
All tests are OK on my fork.

farhi:
I think I'm done, at last... ;-)

willend:
I know I have been 'a bad boy' myself and made similar many-file commits / PR's in the past, but honestly getting an overview and understanding is much easier if we can group things a little better.
(This pr touches 28 files in one go as is and is a bit hard getting the full overview of. ;-) )

willend:
But - merging this time round and we'll figure it out later. But please keep in mind my 0.02€ from above. :)

farhi:
Agreed ! This was massive, and I'm aware of it. Apologies.
Next time I'll create branches on my side to push into smaller bits.

PR #1954, opened 2025-04-22
Add Windows / arm64 in CI (only msvc and sans MPI for now)

willend:


farhi:
Perhaps WSL on Windows/ARM would be the right solution ?

willend:
@farhi sure, will absolutely work. And even the (emulated) "stand alone" x86_64 installer is functional. This is more the point of adding in some very simple tests to check for portability alone. :-) (And wanted to add something since @tkittel advertised that he included this slightly obscure platform in his CI for NCrystal and MCPL.)

farhi:
One other distribution channel is Arch, which is certainly more widely spread by now that Windows/ARM.
The AUR packages are obsolete:
- https://aur.archlinux.org/packages/mcstas
- https://aur.archlinux.org/packages/mcxtrace

willend:
And of course we have good old https://packages.mccode.org/debian/dists/stable/main/binary-arm64/

willend:
Screenshot 2025-04-23 at 12 17 11

tkittel:
@farhi So I think that there is a difference between on one hand distribution channels and packaging, and on the other hand the basic platform support (i.e. do you have problems if some integer type change size or whatnot, or does the compilation break with some compiler in some conditions).

I assume this PR is about the latter: verifying that McStas code works on each commit on Windows/ARM in whatever environment github provides for that platform. That way, we can quickly catch basic errors which would be hard to track down later after a new release is out and installed through some specific distribution channel.

Of course, ideally, we would _also_ check packages in all distribution channels thoroughly, but some fruits are hanging higher on the tree than others ;-)

willend:
The default is in fact to use `epoch` as seed. ;-)

willend:
... that being said, there _are_ apparently minor differences still...:
Screenshot 2025-04-23 at 13 40 26

willend:
I will now try giving our 'conda proper' / MSVC implementation a go also...

willend:
MSVC implementation agrees numerically with "mingw-solution". To me it seems evident that this is down to the platform specific implementation of our `unsigned long randstate[7];` ...  `#include <inttypes.h>` and then `uint64_t` is probably what it should in fact be...?

farhi:
Of course, Debian is not Windows, but regarding architecture support, we have:

- https://packages.debian.org/unstable/mcstas

which tests, for free, much more that what we are discussing here. The issue here is only Windows and its available conda libraries, not really architecture/CPU.

tkittel:
@farhi, the more testing the better :-)

What tests are running in the debian system actually? Peters full suite of instrument tests, or just a few selected ones like in the github CI?

Personally, although more testing is better, I find the most value in having tests running on each commit in a PR. 

Also, as concerns the MSVC compiler for all its faults seems to be much better than clang/gcc at reporting suspicious usage of integers. Or perhaps it is not better, but it did uncover a lof of previously uncovered bugs when I started to build NCrystal and MCPL with MSVC. But admittedly I did not find a lot of stuff when also adding windows-arm into the mix. However, I feel better about having it tested in CI before some person shows up at a workshop with a windows arm laptop and runs into issues :-)

willend:
We (sort of) have, except not tested at this level / between platforms.

willend:
@farhi and @tkittel for what it is worth, the discussion here and the addition of CI running on Windows arm already lead to one improvement, namely achieving the same RNG behaviour across platforms :-D
https://github.com/mccode-dev/McCode/pull/1955 
(I think these small platform-tests are really worth.)

PR #1975, opened 2025-04-29
Update vs code extension

Lomholy:
Added new gui like features for the vs code extension.


willend:
@Lomholy thanks, very nice! :-) It unfortunately missed 3.5.27 which I am about to announce came out, but will make it to the next release. :-)

PR #1979, opened 2025-04-29
Mcxtrace fluo fix pow 0

farhi:
**mcxtrace**
A fix in `FluoPowder` which "sometimes" brings diffraction I=0 due to a rounding error on some architectures (amd64 behaves differently than arm64).
Also add a Fluo+SX GROUP example in `Test_SX`.
There are still a few strange behaviours in the intensity for SX with fluorescence...


willend:
@farhi OK, I merge. 

If I find time I might try running e.g. cppcheck code analysis on the instr - compilation with clang or msvc also sometimes bring intersting hints?

PR #1989, opened 2025-05-01
McGui: show current instrument name in dialogues

farhi:
Few commits to:

- show instrument name in main window title
- show instrument in the Run dialogue. I could not set the title itself. Better than nothing.

This way we know what dialog/instr pair we are looking at...


willend:
Good call @farhi

PR #2000, opened 2025-05-04
Fix compilation of McXtrace Test_PowderN instr

willend:
Failed in nightlies, correct parameter naming in instr.


farhi:
Thanks, had not spot this one.

PR #2002, opened 2025-05-07
Monitor_nD pixel id and buffer fix

mads-bertelsen:
Two updates to monitor_nD event handling:

1) Buffer was dependent on ParticleCount, which incremented on all particles reaching the trace function, but it should only be incremented whenever the ray is actually within bounds. This caused many 0's in the output as rows were basically skipped.
2) The pixel id had to be the second last variable to function correctly with the BINS matrix, code was added to find the right column.


willend:
Cool, thanks for that @mads-bertelsen. I will run the "nightlies" based on this branch shortly - worth investigating if anything systematically changes in the "example suite" I think?

willend:
I have convinced myself that we can merge - test looks OK with correct setup and version of NVHPC

PR #2006, opened 2025-05-10
WIP Mccode unify grammar 0

farhi:
This PR is an attempt to unify our code generator.

**observation**
Currently only one line in `files.c` differs between `src/mcstas` and `mcxtrace/src`.
So we could in principle get only one common directory to hold a unified code generator, and use `MCCODE_PROJECT`(1 or 2) to selectively choose the proper component directory structure in `files.c`.

**potential solution**
I guess a solution could be:
- to update `files.c`to handle `MCCODE_PROJECT`
- to unify `mcstas/src`and `mcxtrace/src`into e.g. `common/src`
- to link `mcstas/src`and `mcxtrace/src`as `../common/src`

**this PR**
Is not as above. It just installs a single version of all files in `src` except `files.c`, and link all other `mcxtrace/src`to corresponding `mcstas/src`versions.
The local build is OK under Linux/MacOS, but not Windows, as the new links are not properly handled.
Please feel free to reject this PR, and/or create a new one with appropriate modifications.


willend:
@farhi as discussed earlier I propose we transition to using a common mccode/src area and work there (which should allow us to work around symlink issues on Windows). I will try to find time to look at this later in May.

willend:
Superseded by https://github.com/mccode-dev/McCode/pull/2008

PR #2009, opened 2025-05-11
Unify instr y c py: unify our code generators for C and Python

farhi:
This is an attempt to unify our code generators, now that we have a common McStas/McXtrace cogen both for C and Python. The motivation is to have the grammar defined only once.

The only requirement in this PR is to set a GENERATE_PY define at compile time, which is branched in a few places (e.g. `instrument.y` and `cexp.c`). Otherwise the default is to define a GENERATE_C symbol.

This is achieved when building from `cmake:InstallMCCODE` in which we now have a line:
```
AppendDef(GENERATE_PY=1)
```
just before assembling the bison/flex `pygen` into an executable.


farhi:
OK complete failure !!!
We now have only the python generator... Make Greg happy.

willend:
On macOS `sed: 1: "py-instrument.tab.c": extra characters at the end of p command`, I will have a look. 
If you ar using `-i` @farhi a suffix is required on macOS (`BSD sed` peculiarities I think) .

willend:
My guess is that the test will run through now. I will see if I can come up with another more CMake-esque solution... Better avoid the non-standard sed approach if we can (not so platform independent, think about MSVC on windows...)

willend:
Ha! Much better IMHO

farhi:
Many thanks Peter,

we now have a unique code generator for all solutions, great !
I've also an other PR to enhance the grammar with multiple COPY/EXTEND for all component sections, to allow merging e.g. samples or other complex optics. Will push it soon...

PR #2012, opened 2025-05-11
Grammar multiple comp copy extend

farhi:
This is a PR to allow multiple `%{ %}/COPY/EXTEND` blocks in component definition sections SHARE DECLARE INIT TRACE SAVE FINALLY DISPLAY. 

Up to now, only syntax:
- `%{ %}`
- `COPY(compdef) EXTEND`

where allowed. 

The new syntax allows to mix/alternate as many blocks as required, to e.g. set some C code, append something from a component, possibly an other one, and end with an EXTEND. 

An immediate application will be to shorten a lot the new FluoPowder and FluoCrystal by having e.g.:
```
SHARE COPY(Powder) COPY(Fluorescence)
```
and this syntax could probably be applied as well to the INIT and DECLARE blocks. This way, we may reduce a lot code duplication, without creating dedicated .c/.h libraries.

There is full backward compatibility.

### DIFF
--- a/cmake/Modules/InstallMCCODE.cmake
+++ b/cmake/Modules/InstallMCCODE.cmake
@@ -261,35 +261,27 @@ macro(installMCCODE)
   )
 
 
-  ## Generate lex.yy.c with flex
+  ## Generate lex.yy.c with flex (C generator)
   add_custom_command(
     OUTPUT work/src/lex.yy.c
     COMMAND "${FLEX_EXECUTABLE}" -i "${PROJECT_SOURCE_DIR}/src/instrument.l"
     WORKING_DIRECTORY work/src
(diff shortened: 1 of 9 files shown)


willend:
@farhi thanks this all looks good. 

@g5t FYI multiple heads-up wrt. code generator:
* The structure of mcstas/src  mcxtrace/src is now collapsed in mccode/src and the grammar only defined in one place
* I have allowed instruments compiled without `--trace` to accept `--trace=0` and only fail on a non-zero value
* @farhi has added this PR that allows multiple `COPY` and `EXTEND` in a single comp code

all in all there is probably some follow-up required for the `mccode-antlr` side, we can have a chat about it one of the next days?

PR #2023, opened 2025-05-14
Swap default trace behaviour to 'enabled'

willend:
PR to swap default trace behaviour:
1) Enable --trace by default
2) Add --no-trace to explicitly disable trace
(No need for changes in runtime or tools)

I would like the 0.02$ of you on https://github.com/mccode-dev/McCode/discussions/2011 before any merging
@farhi @g5t @mads-bertelsen @tweber-ill and maybe even @ebknudsen or @tkittel have a point of view? :-)

No direct implications for the "standard end-user" as `mcrun` / `mcgui` / `McStasscript` all use this default already. It only affects the default of the actual code generator that historically needed an explicit `mcstas -t` / `mcxtrace -t`. The `-t` becomes implicit with this PR and a handle to disable (`--no-trace`) has been added.

(For those worrying about added performance penalty - it is kind of artificial to talk about since anyone using the "high level tools" did not use this for years already... But a recent measurement says one may gain maximum 5% in a simulation by explicitly disabling via `--no-trace`.)

### DIFF
--- a/.github/workflows/mcstas-basictest.yml
+++ b/.github/workflows/mcstas-basictest.yml
@@ -342,7 +342,7 @@ jobs:
            #if [ "x$(uname)" == "xDarwin" ]; then export MCSTAS_CC_OVERRIDE=/usr/bin/clang; fi
            if [ "x$(uname)" == "xLinux" ]; then
              ../install_mcstas/bin/${MCRUN_EXECUTABLE} --format=NeXus --IDF templateSANS_Mantid.instr lambda=6
-             ../install_mcstas/bin/${MCRUN_EXECUTABLE} -c --verbose --mpi=2 --format=NeXus --IDF templateSANS_Mantid.instr lambda=6 -d MPIoutput
+             ../install_mcstas/bin/${MCRUN_EXECUTABLE} -c --verbose --mpi=2 --format=NeXus --IDF templateSANS_Mantid.instr lambda=6
            fi
 
     - name: 'Tar output files'
--- a/.github/workflows/mcstas-basictest.yml
+++ b/.github/workflows/mcstas-basictest.yml
@@ -342,7 +342,7 @@ jobs:
            #if [ "x$(uname)" == "xDarwin" ]; then export MCSTAS_CC_OVERRIDE=/usr/bin/clang; fi
            if [ "x$(uname)" == "xLinux" ]; then
              ../install_mcstas/bin/${MCRUN_EXECUTABLE} --format=NeXus --IDF templateSANS_Mantid.instr lambda=6
-             ../install_mcstas/bin/${MCRUN_EXECUTABLE} -c --verbose --mpi=2 --format=NeXus --IDF templateSANS_Mantid.instr lambda=6
+             ../install_mcstas/bin/${MCRUN_EXECUTABLE} -c --verbose --mpi=2 --format=NeXus --IDF templateSANS_Mantid.instr lambda=6 -d MPIoutput
            fi
 
     - name: 'Tar output files'
--- a/mccode/mccode.md.in
+++ b/mccode/mccode.md.in
@@ -8,7 +8,7 @@
 
 # SYNOPSIS
 
-**@FLAVOR@** [-o file] [-I dir1 ...] [-t] [-p] [-v] [--no-main] [--no-runtime] [--verbose] file
+**@FLAVOR@** [-o file] [-I dir1 ...] [-t] [--no-trace] [-p] [-v] [--no-main] [--no-runtime] [--verbose] file
 
 # DESCRIPTION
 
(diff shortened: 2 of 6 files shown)


farhi:
This is low level option: as you say most users will not even notice. 
So any solution (default with/without `--trace`) is good to me. Now, when using the GUI, all compile must use `--trace` ON to allow 'Trace' mode for 3D representation. 
Then the `--no-trace` makes sense for those that wish to gain 5% in speed, from the command line.
So I vote for with:
- `--trace` as default (incl. from GUI)
- `--no-trace` is available at compile time to gain 5% speed.

tweber-ill:
Hi all, yes, that make sense, I also agree with Emmanuel.

willend:
@tweber-ill @farhi I have now added `--no-trace` to `m[c/x]run` - it naturally implies `-c` and in fact suppresses `-t/--trace if this was given. 

The `mccode.md.in` and `mcrun.md.in` pages have also been updated with this new option.

willend:
All seems functional now:

* Plain `mcstas/mcxtrace` run implies `--trace` enabled
* New cogen option `mcstas/mcxtrace --no-trace` to suppress trace
* New `mcrun` option `--no-trace` (implying `-c`) to suppress trace
* Any resulting binary instruments accepts `--trace=0` irrespective of the above, but can only actually run with `mcdisplay` if:
  * support was compiled in
  * `--trace=1/2` is supplied.

willend:
@g5t I'd like your 0.02$ on this if possible? Also, I am adding a PR over at mccode-antlr with (half-baked?) suggestions for similar behaviour there.

g5t:
Through your `--trace=0` flag, do you have an idea of how much of that 5% performance hit is printing to `STDOUT`? 
(I wonder if maybe even a `--quiet` flag that inserts `#define printf noprintf` or similar would be worthwhile, for those of us who've made _very_ verbose components and now can't be bothered to cut down on the extraneous printing)

willend:
I think none of that 5% performance - @g5t the thing is that with `--trace=0`, no actual event data printf happens? 

But let me do another measurement between
1. a selection of instruments with `--trace` capability built in - but normal sim -> datafiles mode
2. the same set of instruments without trace built in -> datafiles mode
(This is the thing I did in a handheld fashion not long ago that gave ~5% penalty of 1. vs 2. The moment one would _actually_ `--trace=value` some stuff the simulations would be waaaaay slower. This we can also measure, but will generate lots of noise :-) )

willend:
And just to make it as clear as I can for everybody watching:

**Historical implementation:**
1. `--trace` flag of code-generator meant "please build in viz capability", actually by a `#define`.
2. the bare `mcstas` command would not raise `--trace` (i.e. by default off)
3. `mcrun` and `mcgui` have added `--trace` to its their mcstas calls for ~ two decades, i.e. somewhat confusing, the opposite of point 1.
4. Before modernised webgl arrived, `--trace` of a generated binary was simply a flag. Set or not. If a binary without compiled-in trace support got the flag it would simply complain.
5. After modernised webgl `mcdisplay` came along, the `--trace` of a resulting binary has taken a value, i.e.:
  * =0 don't run in trace mode
  * =1 use historical multiline visualisation
  * =2 use modernised visualisation with some support for solds
6. Since a few weeks a binary without compiled-in trace support has tolerated `--trace=0`, of course complains if value is non-zero since it can not printf any info about instr / neutrons.

**The new stuff (b and beyond part of PR):**
1. point 6 from above
2. New default for code-generator, bare `mcstas` now means the to build in the `--trace` support, i.e. same default as `mcrun/mcgui` have had for ages
3. New cogen option `--no-trace` to disable trace mode, i.e. set `#define` so that the binary should not get trace option, i.e. corresponding to the old bare `mcstas` command
4. mcrun has a new `--no-trace` option that requests the `--no-trace` option of the code-generator.
5. As compilation without trace is _required_ to get a binary that "does not talk trace", I felt it was natural to couple to `-c` of `mcrun`, i.e. to force-compile.

willend:
(Just realised that since the `--no-trace` option for mcrun is in a way "for specialists" we could also just ignore it unless given tougher with `-c`? "It only costs 5% runtime, and if the binary is already there, don't bother")

willend:
I liked the above thought ^ about not force-compiling enough that this last commit is for that:

`--no-trace` needs to be run "from fresh" or with `-c` to be effective

ebknudsen:
First things first - I have no real objection to making ```--trace``` the default. I mean - you probably don't want it if you're on a headless system (as in comp-cluster), but if so you could make it part of your default config on that system.

That said, I have a slight dislike to the last one, I confess. Lets say you have done a recompile with --no-trace and now want to run with --trace instead. This would require another recompile, no? The only way to ensure this is always clean is to always recompile if either --trace/--no-trace is set. Or am I wrong?

In any case am doing things to spiders that only other spiders should do :-D so feel free to disregard. It really is a bit construed.

willend:
I think the fair assumption is that
* all users (including average ones) will use --trace (via gui or terminal command)
* relatively few users will ever use --no-trace - and those who will can be assumed to know what they are doing :-)

PR #2029, opened 2025-05-17
Mcxtrace fluo add detector 0

farhi:
Move escape + pile-up stuff from Fluo samples to a new monitor `Fluo_detector`. 
This latter is tested within the `Test_Fluorescence`.
Indeed, these feature' are bound to a detector, not a sample scattering process.


farhi:
That's all good, I merge...

PR #2037, opened 2025-05-21
DO NOT MERGE (Attempts to stabilize fluo)

willend:
@farhi it is a bit weird, running with identical seeds the fluo signal in SOLEIL_LUCIA is fluctuating up to an order of magnitude...

Transmission seems mostly OK.

On this branch I have tested with [cppcheck](https://cppcheck.sourceforge.io) and aligned almost all hints I got in direction of uninitialised vars etc., have also experimented with the `message` system from `XRL` - still didn't nail it.

Here is the 1. output from Fluo_detector and 100 times the same sim / seed...: 
Screenshot 2025-05-21 at 16 05 06

### DIFF
--- a/common/lib/share/read_table-lib.c
+++ b/common/lib/share/read_table-lib.c
@@ -427,6 +427,10 @@ void *Table_File_List_store(t_Table *tab){
       double *dataf;
       s     = (float*)data;
       dataf = (double*)malloc(sizeof(double)*nelements);
+      if (!dataf) {
+	fprintf(stderr, "Could not allocate data block of size %i\n", nelements);
+	exit(-1);
+      }
       for (i=0; i<nelements; i++)
         dataf[i]=s[i];
       free(data);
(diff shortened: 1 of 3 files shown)


farhi:
I agree, this weird. I could already find some variable results during my own tests, and decided it would do for this round.
But this is clearly not satisfactory.

willend:
If you want we can also simply test on the value of the transmission monitor for now? And write an issue reminding us to do a deeper dive later?

willend:
Otherwise my thoughts go in the direction of forcing independent use of each of the "fluo branches" one at a time - to potentially zoom in on the problem...

farhi:
Hi Peter, thanks for all the testing and code analysis. Needs more thinking...
OK to test the transmission monitor.

farhi:
I'm working on this at:
https://github.com/farhi/McCode/tree/4-mcxtrace-fluo-fix-again

There is also an issue with SPLIT.
I'll nail down the problem and cut it into pieces "ah ah AH AHHHH"

farhi:
The issue possibly arises from `rand01` calls in the `SHARE` section of fluo components.
The rand01 requires an implicit definition to a `randstate` in `_particle` but these are not defined in SHARE.
The resulting random numbers are ... well ... 'very' random.

willend:
@farhi it might be possible to get a random number via direct use of mt_random which should still be in our code base?

Alternative we might consider to add the _particle definition (and a single particle pr. mpi process?) already in / before init ?

farhi:
I have fixed the `rand01` issue in SHARE. All random numbers are generated in the TRACE and passed to SHARE functions. 

- :+1: The calculation is now stable with fixed seed.
- :-1: But it does NOT fix the LUCIA variability of calculations (with unset seed from clock). 
- :-1: Still has issues with the use of SPLIT for Fluorescence samples.

I'll investigate further. There is a slow improvement :smile:

farhi:
Your commits are probably OK to push. My as well.
But this is not enough yet.
I have created the issue #2044 to track our thoughts and attempts.

PR #2042, opened 2025-05-24
Add exit attenuation to Incoherent.comp for finite order scattering

Lomholy:
[inst_bash.zip](https://github.com/user-attachments/files/20427436/inst_bash.zip)
Tests comparing Incoherent single scattering to analytical results, showed that Incoherent.comp was scattering with too high an intensity at order=1, but at order=0 it was lower.

Attenuation was not added on the way out of the sample, after scattering, resulting in this error.

With this fix, the analytical and simulated results match.

A bash script, and an instrument which highlight this is attached. The predicted scattering intensity at the monitor is 27.12 n/s, and this is achieved with this addition. The bash script should be run with "thin" as an argument
[inst_bash.zip](https://github.com/user-attachments/files/20427439/inst_bash.zip)

### DIFF
--- a/mcstas-comps/contrib/Monochromator_bent.comp
+++ b/mcstas-comps/contrib/Monochromator_bent.comp
@@ -1,16 +1,18 @@
-/******************************************************************************
+/*******************************************************************************
 *
 * McStas, neutron ray-tracing package
 *         Copyright 1997-2002, All rights reserved
 *         Risoe National Laboratory, Roskilde, Denmark
 *         Institut Laue Langevin, Grenoble, France
 *
-* Component: Monochromator_bent
+* Component: Monochromator
 *
 * %I
-* Written by: Daniel Lomholt Christensen <dlc@math.ku.dk> with help from Jan Šaroun
+* Written by: Daniel Lomholt Christensen <dlc@math.ku.dk> 
+* with help from Jan Šaroun
 * Based on the model implemented by Jan Šaroun in the paper published in
-* Nuclear Instruments and Methods in Physics Research Section A: Accelerators, Spectrometers, Detectors and Associated Equipment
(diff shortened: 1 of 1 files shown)


willend:
@farhi confirmed that the fixes here seem to rectify `SOLEIL_LUCIA`:
Screenshot 2025-05-26 at 09 47 13
Screenshot 2025-05-26 at 09 47 27

Lomholy:
Hi Peter,

Yes of course. Here is one:

$$ 
\frac{dN_{\rm real}}{dt}= \Psi A \rho_{\rm c}\sigma_{\rm inc}
                   \frac{\Delta \Omega}{4 \pi} a_{\rm lin}.
$$

Where $\Psi=1e7$, $A=1$, $\rho=0.0723$, $\sigma_{\rm inc}=\sigma_{\rm abs} = 5.08$, $\frac{\Delta \Omega}{4 \pi} = 10/99.95^2$ and $a_{\rm lin}= \exp(-(\sigma_{\rm inc} + \sigma_{\rm abs}\lambda/1.7982)\rho l_{avg})$


And here is a comparison figure for the backscattering:

image

farhi:
The failing tests are related to W$ 2019 which is a deprecated pipeline. I guess it's actually OK.

willend:
Indeed @farhi! I had simply forgotten to merge https://github.com/mccode-dev/McCode/pull/2055 that disables Windows-2019. :-)

I am however still waiting for @Lomholy to have a quick last look / update to the instr header.

willend:
@Lomholy I’ve updated this branch against main and e.g. the new LaTeX feature for mcdoc is also present here.

willend:
@Lomholy looks good, will merge for inclusion in 3.5.30 (and will test / fix on GPU if I find the time).

Issue #2044, opened 2025-05-25
BUG: mcxtrace: fluorescence


farhi:
Weird behavior of fluorescence samples for McXtrace.

- Unstable `SOLEIL_LUCIA`
- SPLIT does not reproduce non-SPLIT results

Two attempts to solve have been started:
- https://github.com/mccode-dev/McCode/pull/2037
- https://github.com/farhi/McCode/tree/4-mcxtrace-fluo-fix-again

which do improve results, but are not sufficient.

farhi:
Just pushed a new set of example test values fixes in https://github.com/farhi/McCode/tree/7-mcxtrace-fix-examples
Also made some tests faster to execute so that our test suite can be obtained in less time/cpu.

willend:
I think this was fixed

PR #2045, opened 2025-05-25
4 mcxtrace fluo fix again

farhi:
Attempts to solve #2044

### DIFF
--- a/mcxtrace-comps/data/LiNbO3.cif
+++ b/mcxtrace-comps/data/LiNbO3.cif
@@ -1,0 +1,77 @@
+#------------------------------------------------------------------------------
+#$Date: 2016-02-13 21:28:24 +0200 (Sat, 13 Feb 2016) $
+#$Revision: 176429 $
+#$URL: svn://www.crystallography.net/cod/cif/1/52/17/1521772.cif $
+#------------------------------------------------------------------------------
+#
+# This file is available in the Crystallography Open Database (COD),
+# http://www.crystallography.net/
+#
+# All data on this site have been placed in the public domain by the
+# contributors.
+#
+data_1521772
+loop_
+_publ_author_name
+'Postnikov, A.V.'
+'Caciuc, V.'
+'Borstel, G.'
+_publ_section_title
+;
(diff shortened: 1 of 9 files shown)


willend:
(I might cherry-pick a few of the changes from https://github.com/mccode-dev/McCode/pull/2037 for inclusion here.)

willend:
Are you OK with me pulling in the changes made in https://github.com/mccode-dev/McCode/pull/2037 directly here in this PR? (Then let's keep this PR open for now for your incoming changes also?)

willend:
@farhi FYI I will be on vacation until monday. :-) 

My plan is to release https://github.com/mccode-dev/McCode/releases/tag/untagged-d6551b6607add66a162f (currently in draft) afterwards, the dedicated branch https://github.com/mccode-dev/McCode/tree/prep-for-3.5.30 contains most of the required preparation. 

So with a little luck this next release will contain lots of goodies

farhi:
I'm still struggling with the issues in #2044.
I have fixed a number of oddities, but:
- the main instability on the final flux with random seed remains.
- the SPLIT changes results wrt non-SPLIT.

farhi:
I think there is some black magic in the background. I'm investigating. 
The random number stuff is fixed, but I think some dark side influence is still sneaking into the calculation to make the flux very unstable. 
BTW, the flux is also unstable at the `sample _location`, before the `Fluorescence` component...

willend:
@farhi then an obvious candidate is the contributed `Bragg_crystal` - common denominator to `DBD_IBM_Si_analyzer_BC`: also acting up on the [nightlies](https://new-nightly.mcxtrace.org/todays-datafiles/2025-05-29_output.html)!

willend:
@farhi sounds like a good idea! As I remember there are two variants of that Bragg crystal comp in  `DBD_IBM_Si_analyzer_BC`, so if you correct / experiment there probably both need a look.

farhi:
I think we can merge this part.
I'll work on the `Bragg_crystal` in a separate branch.

PR #2046, opened 2025-05-27
Update MCPL components for MCPL 2.2.0 features (including stat:sum).

### LINKED ISSUE #2018: Update MCPL components to make files merge-able and retain correct flux information
Once MCPL 2.1.0 is released with the new support for additively mergeable statistics parameters (cf. https://mctools.github.io/mcpl/format_statsum/, https://github.com/mctools/mcpl/pull/89) we should update all (7?) MCPL input and output components in McStas and McXtrace to use this to make sure we can merge output files from repeated runs and retain correct particle weights and knowledge of overall initial rays simulated through the instruments.

The context is that most(?) source components in mccode tend to divide the initial particle weights with the total number of particles simulated (N). So as it is, MCPL files from jobs with respectively N1 and N2 initial particles can not readily be combined, without custom code rescaling all the individual particle weights.

After a discussion with @willend and @g5t about the best way forward, we agreed on the following long-term goal:

1. Add support for additively merge-able statistics parameters in MCPL headers (this is almost done).
2. Make sure that MCPL particle weights written from mccode components will have weights p*N rather than just p, effectively undoing the initial division with N in the source.
3. Record the value of N in a new MCPL `stat:sum` entry using the key `initial_ray_count`. The key name was chosen to not refer to neutrons in McStas exclusively, to potentially make it easier to share tools between McStas, McXtrace and other instrument simulation applications.
4. Modify MCPL input components to read the `initial_ray_count` value (N) and divide particle weights by it, `p=mcpl_weight/N`.

As a migration/transition strategy, the MCPL output components will get a new `weight_mode` flag, which users **for a while** can use to pick the previous behaviour. Initially it will default to the old behaviour, and anyone not picking the new behaviour will get a deprecation warning, instructing them that the old behaviour will eventually become unsupported (and redirecting them to a documentation page for further information).

We also briefly discussed a related issue, that ssw2mcpl should encode the MCNP (/phits/openmc/...) statistics as well in some `stat:sum` parameter, allowing the MCPL_input component to perform a proper scaling.

tkittel:
Update MCPL components for MCPL 2.2.0 features (including stat:sum).

This PR aims to update the various MCPL components to take advantage of latest features, namely:

* Use stat:sum (cf. #2018) to allow retention of statistics in file merges.
* Use new MCPL utility functions to simplify MPI creation and merging of output files.
* A bit of general cleanup.

Fixes #2018

### DIFF
--- a/.github/workflows/mcxtrace-winbasictest-mingw.yml
+++ b/.github/workflows/mcxtrace-winbasictest-mingw.yml
@@ -112,7 +112,7 @@ jobs:
       working-directory: run_mcpl_input
       run: |
         copy ..\install\share\mcxtrace\resources\examples\Tests_MCPL_etc\Test_MCPL_input\Test_MCPL_input.instr test.instr
-        copy ..\install\share\mcxtrace\resources\examples\Tests_MCPL_etc\Test_MCPL_input\voutput.mcpl.gz test.instr
+        copy ..\install\share\mcxtrace\resources\examples\Tests_MCPL_etc\Test_MCPL_input\voutput.mcpl.gz .
 
     - name: Prepare MCPL_output instrument
       working-directory: run_mcpl_output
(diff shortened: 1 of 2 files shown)


willend:
All MCPL instruments check out in McXtrace, single core, mpi and openacc:
Screenshot 2025-05-27 at 14 55 58

All MCPL instruments check out in McStas, single core, mpi and openacc:
Screenshot 2025-05-27 at 15 19 43

willend:
It seems the conda envs on the GPU box are ancient and may require replacement / so nightlies might fail on GPU until I get around to fix that ...

PR #2051, opened 2025-06-01
mcxtrace: fix test values after fluo update

farhi:
I fix test values after the fluorescence sample component update.


farhi:
I see. There is something really odd: the compilation time is much higher as before.
Many examples are now failing. Will have a look in the train tomorrow, while going to Bordeaux.
I have the feeling this is a change in a core library...
Is there a way to get the log of the calculations ? 
The stdout is only given when compilation fails.

PR #2062, opened 2025-06-11
Awaiting reviews. Grammar change, use INHERIT instead of COPY in comps

willend:
Here is the first footwork toward what was discussed in https://github.com/mccode-dev/McCode/discussions/2041 

Seems partially functional, try with the attached comp + instr and McStas:
[BNL_H8.instr.txt](https://github.com/user-attachments/files/20691902/BNL_H8.instr.txt)
[Diaphragm.comp.txt](https://github.com/user-attachments/files/20691903/Diaphragm.comp.txt)

Comp corrections are incoming.

### DIFF
--- a/doc/GRAMMAR/ADR-records/ADR_20250612_INHERIT_COMP.md
+++ b/doc/GRAMMAR/ADR-records/ADR_20250612_INHERIT_COMP.md
@@ -32,6 +32,9 @@ In a zoom meeting on June 11th 2025 it was agreed between  @farhi,
 * User home-grown McStas or McXtrace components will break and need
   migration to `INHERIT`. It is believed that very few - if any - such
   components exist in the wild. 
+* Important: `INHERIT` does not ensure that component setting
+  parameters used in copied section(s) are copied too. (This to allow
+  complete modification of the parameter interface for a "derived" component.) 
 * CHANGELOG should be crystal clear on this change / potential incompatibility
 
 ## Behaviour


willend:
Now running "testsuite" runs with McStas and McXtrace to get a full picture.

mcstas-comps/optics/Diaphragm.comp
farhi, line 1: I think you just can leave a single line
```
DEFINE COMPONENT Diaphragm INHERIT Slit
```
which also copies all other parts if I remember well).

mcxtrace-comps/optics/Diaphragm.comp
farhi, line 1: Same as for McStas, a single line
```
DEFINE COMPONENT Diaphragm INHERIT Slit

END
```
should do...

willend:
@farhi I agree but will give a chance for @g5t or @mads-bertelsen to put their 0.02€ on what was implemented. :-)

mccode/src/instrument.y
```diff
@@ -306,25 +307,25 @@ comp_share: /* empty */
...
       }
-    | "SHARE" comp_share_copy_extend
+    | "SHARE" comp_share_inherit_extend
       {
         $$ = $2;
       }
 ;
 
-comp_share_copy_extend: /* empty */
+comp_share_inherit_extend: /* empty */
```
g5t, line 324: This matches
- `SHARE`
- `SHARE %{...%}`
- `SHARE EXTEND %{...%}`
making the `EXTEND` keyword superfluous. 

Explicitly making the keyword optional (by ignoring its token, in the same way that comments are ignored) would simplify this repeated set of rules.

g5t (changes requested):
It _looks_ like the new component-definition section replication method, including replicating from multiple components and prepending or appending extra code blocks in the process, has been successfully renamed from `COPY` to `INHERIT`.
The fact that tests pass with the associated requisite change in test instrument/component files is positive.

My two comments on `instrument.y` should probably be addressed in separate issues.

What's missing, however, is an Architectural Decision Record (or similar) in the code/repo describing what problem this change addresses, why this particular solution was chosen to address the problem, what the intended new grammar syntax is, etc.

mccode/src/instrument.y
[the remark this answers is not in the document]
willend: Added https://github.com/mccode-dev/McCode/issues/2065 to fix this later

willend:
@g5t I've added a first-of-a-kind ADR entry in `McCode/doc/GRAMMAR-ADR-records`, now part of main and PR-associated record is `ADR_20250612_INHERIT_COMP.md`

doc/GRAMMAR-ADR-records/ADR_20250612_INHERIT_COMP.md
```diff
@@ -0,0 +1,35 @@
...
+## Consequences
+
+* Syntax becomes clearer - either `COPY(reference)` with parenthesis in
+  instruments or `INHERIT reference` in components.
+* Existing components with use of `COPY` must be migrated (already
+  done for repo comps in PR.
+* User home-grown McStas or McXtrace components will break and need
+  migration to `INHERIT`. It is believed that very few - if any - such
+  components exist in the wild. 
+* CHANGELOG should be crystal clear on this change / potential incompatibility
```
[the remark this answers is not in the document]
willend: To me it in a way makes sense to NOT automatically copy setting parameters. Think of the situation where a user wants to derive a component from another one, "suppressing"/"replacing" some functionality?

PR #2074, opened 2025-06-16
[Add] repository dispatch action to notify tag changes in registries

g5t:
The added workflow will only run on creation or deletion of branches and tags, and will be skipped for branches.
If a tag is pushed or deleted a message is sent to [mccode-pooch-registries](https://github.com/mccode-dev/mccode-pooch-registries) which causes a workflow action there to update its tagged registry files.

This should allow for closer synchronization between `McCode` releases and `mccode-antlr`.


willend:
@g5t sounds good, merging that in.

PR #2081, opened 2025-07-09
Add in support for splitting a scan into many cpu processes

Lomholy:
Adds --scan_split keyword to mcrun, which splits a scan into many different cpu processes.

This allows optimised run speed of scans where the individual run time is not the bottle neck, but rather the starting and stopping of the run is the bottle neck.


willend:
@Lomholy in addition to the documentation-stuff from above, I have now added a little code to
* ensure a mccode.sim is produced in each "scan split" subfolder (like in the normal scan) 
* added some heuristics to shift the RNG seed pr. thread process to avoid bias

Merging now

PR #2083, opened 2025-07-23
Update SANS_Spheres to align with analytical validations.

Lomholy:
Absorption was a factor of 100 larger than is postulated in the description and in the code.

It used the wrong distance travelled for calculating the absorption.

Both of these things have been fixed. With this, it aligns with my analytical validation.


willend:
@Lomholy I am now back from vacation. :) 

Thanks for the updated %Example value also. As a precaution and out of old habit I just ran the "test tools" on the modified branch for all the instruments including Sans_spheres (plus a couple of more in fact). It all checks out OK:
https://tmp.mcstas.org/mcstas-test-Sans_spheres/

Merging now.

Issue #2084, opened 2025-07-30
(Automated) Conda builds fail on PPC Linux due to use of -mtune=native


willend:
Automated conda builds for Linux PPC fail in final McStas funktionaly test on Linux ppc:

```
cc1: error: unrecognized argument in option '-mtune=native'
cc1: note: valid arguments to '-mtune=' are: 401 403 405 405fp 440 440fp 464 464fp 476 476fp 505 601 602 603 603e 604 604e 620 630 740 7400 7450 750 801 821 823 8540 8548 860 970 G3 G4 G5 a2 cell e300c2 e300c3 e500mc e500mc64 e5500 e6500 ec603e power10 power11 power3 power4 power5 power5+ power6 power6x power7 power8 power9 powerpc powerpc64 powerpc64le rs64 titan
```

To fix, `McCode` requires patching. At this point not completely sure if the `-mtune` flag is added by ourselves (`cmake/Modules/PlatformDefaults.cmake`) or if it arrives via conda `CFLAGS` or `LDFLAGS`

Via https://github.com/conda-forge/mcstas-suite-feedstock/pull/165

willend:
It in fact looks like '-mtune=native' is from our own "hard-coded" list in PlatformDefaults.

willend:
Fixed for mccode code base yet some dependencies are missing to make mcstas/mcxtrace fly on ppc64le

PR #2085, opened 2025-08-04
Fix for ppc64 Linux (via conda)

willend:
Remove -mtune=native as (e.g. conda) usually has sets a reasonable local default (e.g. -mtune=haswell)

Should eventually rectify #2084


willend:
Dependences missing, so ppc64le is disabled again.

PR #2094, opened 2025-08-14
Allow Phonon_simple to accept more than 1 solution

Lomholy:
Allows Phonon_simple to use more than 1 solution, by removing an if statement, and extending the vf_list to be as long as the new input parameters, that allow for finetuning scan steps below and above the elastic line.

Also change the parms array to be a struct, containing names of the parameters.


willend:
Once I patch your version of the code in the direction indicated above, the code compiles on GPU but crashes... (Still very likely a thread-collision issue...)

I will likely start from fresh on a new branch and implement your changes one by one / add safety belts on the way... Stay tuned.

Lomholy:
I can confirm your observation. Must've missed that one 😅

willend:
OK @Lomholy then I will proceed to merge. I have on my list to look at the Bent mono also in respect to the GPU side.

PR #2098, opened 2025-08-25
Add in order parameter to make it possibly to use no multiple scattering

Lomholy:
Previously my_s would contain both my_inc and my_s_v2_sum, which seems to be an inclusion of all the other possible scattering processes in a sorta mimicking multiple scattering way. 

This however is not in accordance with how the analytical calculations are derived, and this change allows a user to recreate the results one would expect from single scattering.


willend:
@Lomholy thank you for the PR, which I hereby accept and will merge. I attach a the output of a little testing done via my local installation (copied the modified comp to `$MCSTAS/samples` and ran the `mctest` / `mcviewtest` tools) :

```
mctest --instr=PSI_DMC --testdir=$PWD --mpi=auto -n5e7
mctest --instr=ISIS_GEM --testdir=$PWD --mpi=auto -n5e7
mctest --instr=Test_Powder --testdir=$PWD --mpi=auto -n5e7
mcviewtest $PWD
```
Screenshot 2025-08-26 at 11 56 39

PR #2101, opened 2025-09-02
Automatic CI test trigger with -1 return code in case of errors

willend:
Deliberately leaving BNL_H8_simple with compile error initially


willend:
Screenshot 2025-09-02 at 14 25 46

PR #2281, opened 2025-12-18
Allow mcrun / mcdisplay to run / visualize without access to .c/.instr

willend:
### Free-form text area

@g5t with this minor set of changes to mcrun.py / mccode.py, the mcrun and mcdisplay tools can produce simulations or visualisation can proceed without requiring access to underlying .c or .instr file.

To use, simply give the path of wanted binary instead of the .instr file, i.e.:
`mcdisplay-cad BNL_H8.out -y -n1`

--------------

--------------


willend:
Since everything finally checks out (including a [test-suite run](https://github.com/willend/McCode/actions/runs/20359012594) I am merging this.

PR #2304, opened 2026-02-04
Make backprop thread safe in McStas

willend:
### Free-form text area

This is a candidate solution for particle-scope allow_backprop in McStas which is tested to render ALLOW_BACKPROP thread-safe on GPU (see also https://github.com/mccode-dev/McCode/pull/2303)

--------------

--------------


willend:
Behaviour on current main:
Screenshot 2026-02-04 at 16 04 11


Patch from this PR Tested functional on GPU:
Screenshot 2026-02-04 at 16 00 27

PR #2371, opened 2026-03-25
First attempt at making PowderN load NCMAT data directly via NCrystal

tkittel:
### Free-form text area

In consultation with @willend, this PR adds the capability for the PowderN component to accept NCrystal cfg-strings directly in the reflections parameter. When this is the case, NCrystal will be used during `INITIALIZE` to load the reflection list (as well as `V_0`, `sigma_a`, `sigma_i`, `at_nb`, ...).

So for instance, instead of `reflections="Al.laz"`, one can now write `reflections="Al_sg225.ncmat"`, or `reflections="Al_sg225.ncmat;temp=200K;dcutoff=0.5"` or whatever else is supported in NCrystal cfg-strings. Note that the logic for detecting the reflections string as associated with NCrystal is that any string containing either the substring `".ncmat"` or the substring `"::"` is an NCrystal cfg-string. We might want to fine-tune this to e.g. allow a .laz file to be loaded with NCrystal for debugging. Although one could write `"relpath::Al.laz"` in this scheme to get that effect, it might be a bit too obscure.

The idea is of course is that this makes it possible to leverage NCrystal as a source of materials also in cases where unique features of PowderN are needed (e.g. line broadening or GPU), and where features like the inelastic components of NCrystal are less important.

Other changes to the component:
* I removed the `double Epsilon` field from the `line_data` struct, as it did not seem to do anything.
* During testing of this PR, it came to my attention that the `ncrystal_ncmat2hkl` encoded values for `sigma_i` and `sigma_a` in barns/unitcell while PowderN interprets them as barns/atom. Ultimately this gives an error of a factor of natoms/unitcell. This will be fixed on the NCrystal side (see https://github.com/mctools/ncrystal/issues/337), but the present PR also tries to make PowderN read such files correctly (irrespectively of whether the files were produced with the old or the future `ncrystal_ncmat2hkl` version).

--------------
### Development OS / boundary conditions
_Please describe what OS you developed and tested your additions on, and if any special dependencies are required:_

--------------
# PR Checklist for contributing to McStas/McXtrace
## For a coherent and useful contribution to McStas/McXtrace, please fill in _relevant parts_ of the checklist:
* ### My contribution includes patches to an **existing component** file
  * [x] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the component (please attach as screenshot in comments!)
  * [x] I have ensured that basic use of the component is OK (e.g. an instrument using it compiles?)
[...]

PR #2410, opened 2026-04-22
Dummy commit in PowderN.comp

tkittel:
This is a dummy PR intended to check on the issues observed in #2371


willend:
@tkittel I am closing this one as we (for now) concluded work on PowderN with NCrystal cfg-string support

PR #2535, opened 2026-06-29
Provide SCATTER_func for libraries to replicate SCATTER functionality

mads-bertelsen:
### Free-form text area
The SCATTER keyword works in components and provides the current position of the ray to plotting.

That makro is however not available within .c and .h files in share, so components that do most work there can not visualize the ray path in mcdisplay. Here a small function that replicate SCATTER is added to McStas and McXtrace. It is used in conic.h to show the ray path within nested mirror optics.

The function is called SCATTER_func and takes a pointer to the particle, _particle. It returns void.

--------------
OS X

--------------
  * [X] I am have added reasoning and documentation for the change below
  * [] I am attaching test output in the comments

--------------

### DIFF
--- a/mccode/nlib/share/mcstas-r.c
+++ b/mccode/nlib/share/mcstas-r.c
@@ -108,6 +108,20 @@ _class_particle mcgetstate(_class_particle mcneutron, double *x, double *y, doub
   return(mcneutron);
 } /* mcgetstate */
 
+/*******************************************************************************
+* SCATTER_func: provides function to SCATTER from within libaries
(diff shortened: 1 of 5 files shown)


willend:
Nice! I thought you mentioned something about a similar solution in use in ESS_butterfly-related share code, yet I don’t seem to find anything like that now?

mads-bertelsen:
The library used in the butterfly component just had mcdisplay function calls, making me realize we could do things like this

willend:
OK. Then I will merge like this.

PR #2568, opened 2026-07-09
mcviewtest: spawn the new mcplotdiff-html plotter

willend:
### Free-form text area

--------------

Claude, prompt:

"Now I'd like you to focus on the mcviewtest tool which is part of tools/Python/mctest/.

The mctest tool can generate a set of test data for McStas, based on %Example lines in the instr files. As part of this, it calls the mcplot-html plotter per test result.

The folder structure from an mctest can be found at e.g. http://new-nightly.mcstas.org/todays-datafiles/mcstas-3.99.99_mpi_x_20_1e7_Linux_20260709_0902_23/ 

 Afterwards, mcviewtest is called to generate an overview table - including columns that correspond e.g. to CPU and GPU execution - or different operating systems, see the html file at http://new-nightly.mcstas.org/todays-datafiles/2026-07-09_output.html (spans the above mctest output +  2 other mctest's.

Consider the first column in the html file the 'reference' test. Comparing each of the OTHER columns to the first one,  I'd like you to add 'difference' cells corresponding to a mcplotdiff-html comparion between the dataset outputs, e.g. McStas_8GPU_openacc_mpi_x_7_1e7_Linux_20260709_0001_06_1e7/BNL_H8/1/ and mcstas-3.99.99_mpi_x_20_1e7_Linux_20260709_0902_23/BNL_H8/1/ - adding a link the the mcviewtest template."

Various local adaptation added 

--------------

--------------

--------------


willend:
A first demo of what this can do is available through http://new-nightly.mcstas.org/demo-diffplots/2026-07-09_output.html. Where possible (simulation worked and produced ascii output), the other datasets are compared against the left-most column (“ref”).

While this still has rough edges (e.g. very long legends on links etc.) I think it will be very useful for instantly getting a good idea of what is up when a test starts to fail numerically.

willend:
With the latest few commits (at 5ac3e2c) the output got a couple of stylistic changes.

Overview:
Screenshot 2026-07-10 at 11 03 06
(DIFF separated from rest of cell with a \<hr\>)

Diffplot:
Screenshot 2026-07-10 at 11 03 13
(Datasets labeled ‘A’ and ‘B’ in header, referral to these in link labels)

willend:
After last commit, individual ‘monitor’ plots look like this:
Screenshot 2026-07-10 at 11 18 40
Screenshot 2026-07-10 at 11 18 29
(with link to the underlying ascii datafiles)

where as the ‘diff-monitor’ plots look like this:
Screenshot 2026-07-10 at 11 19 46
Screenshot 2026-07-10 at 11 19 56
(the ascii data do not exist in ‘physical form’)

willend:
Here is a link to the output from mcstas-conda-testsuite as run from the PR bramch
http://new-nightly.mcstas.org/combined/combined_output.html

.github/workflows/mcstas-conda-testsuite.yml
```diff
@@ -133,3 +133,119 @@ jobs:
...
+        environment-file: dependencies.yml
+
+    - name: Configure build and install mcstas
+      id: mcstas-install
+      run: |
+           # Only the Python tooling (mcviewtest, mcplotdiff-html, mcplot-html) is
+           # actually needed here - the diff/report generation is platform-agnostic
+           # regardless of which OS originally produced the monitor data - but we
+           # reuse the exact same install path as the build job for consistency,
+           # rather than maintaining a second, lighter-weight install mechanism.
```
mads-bertelsen, line 169: This is okay for now but could obviously be improved! Maybe we should make an issue on this to remember. Maybe its not a significant amount of time?

tools/Python/mcplot/html/mcplotdiff.py
```diff
@@ -335,28 +352,30 @@ def write_index(outdir, entries, label_a, label_b):
...
+                outfile.write(f"<a href='{a_log}' target=_blank>[ A (log) ]</a>\n")
             if b_lin:
-                outfile.write(f"<a href='{b_lin}' target=_blank>[ {label_b} ]</a>\n")
+                outfile.write(f"<a href='{b_lin}' target=_blank>[ B ]</a>\n")
             if b_log:
-                outfile.write(f"<a href='{b_log}' target=_blank>[ {label_b} (log) ]</a>\n")
-
+                outfile.write(f"<a href='{b_log}' target=_blank>[ B (log) ]</a>\n")
+            if a_lin or a_log or b_lin or b_log:
+                outfile.write("<span class='origlinks'>\n")
```
mads-bertelsen: This should be closing the span, instead it starts a new one

tools/Python/mcplot/html/mcplotdiff.py
```diff
@@ -318,7 +334,8 @@ def write_index(outdir, entries, label_a, label_b):
         outfile.write("  #sizecontrol input[type=range] { vertical-align: middle; margin: 0 8px; }\n")
         outfile.write("</style>\n")
         outfile.write("</head><body>\n")
-        outfile.write(f"<h1>Difference plots: {label_a} &minus; {label_b}</h1>\n")
+        outfile.write("<h1>Difference plots:</h1>\n")
+        outfile.write(f"<strong><ol type=\"A\"><li>{label_a} vs.<li>{label_b}</ol></strong>")
```
mads-bertelsen: This line is different from the previewed webpage in the PR, it lacks the closing parts of the <li> so it doesn't seem to be valid html. Why use a list within an ordered list to show the heading?

tools/Python/mccodelib/mcplotloader.py
```diff
@@ -13,6 +14,47 @@
...
+        block with inconsistent column counts between rows. This keeps the
+        common case fast without sacrificing correctness on the edge cases
+        the original loop happened to tolerate. '''
+    if not lines_block:
+        return []
+    try:
+        ncols = len(lines_block[0].split())
+        if ncols == 0:
+            raise ValueError('empty first line in data block')
+        flat = np.fromstring(' '.join(lines_block), dtype=float, sep=' ')
```
mads-bertelsen: This can create quite a big object in memory before its passed to numpy that makes it into floats. It could be done in parts to avoid having double the text size in memory at once, for example

flat = np.concatenate([np.fromstring(line, dtype=float, sep=' ') for line in lines_block])

mads-bertelsen (changes requested):
Great new capability! Only critical issue is the syntax issue in the generated html that doesn't close the span but instead opens a new one.

.github/workflows/mcstas-conda-testsuite.yml
willend: The comment is honestly directly from Claude and IMHO an over-analysis.

The installation takes max. a minute, I see no reason to set up an independent path, better maintain a standard approach.

tools/Python/mccodelib/mcplotloader.py
willend: I take your word for it. :-D Patch incoming.

tools/Python/mcplot/html/mcplotdiff.py
willend: Copy-paste gone wrong. Patch incoming.

tools/Python/mcplot/html/mcplotdiff.py
willend: Not starting a list within an ordered list, `<li>` means “list item” and are used acrooss any type of `<ul>` and `<ol> ` to identify items in the list. But they are rightfully missing their `</li>`, a little sloppy. Patch incoming.

willend:
@mads-bertelsen I have taken into account most of your comments above - resolved all.

Also, it seems I had further, non-pushed edits on my local checkout, so the last set of commits above became a bit more voluminous than intended, sorry about that.

(I guess the merge-blocking then logically means / requires a re-review?)

willend:
Further updating against main with the purpose of running the updated ‘testsuite’. Will pop a link here after.

willend:
Updated set of ‘testsuite’ runs at http://new-nightly.mcstas.org/combined/combined_output.html

mads-bertelsen:
The commit that wasn't pushed explains some confusion we had over small mismatches between the code that generated the HTML and the HTML in the example page! Now I see what you wanted to achieve with the ordered list and agree it is easier to separate the two long names. Looking through the code again.

willend:
> The commit that wasn't pushed explains some confusion we had over small mismatches between the code that generated the HTML and the HTML in the example page! Now I see what you wanted to achieve with the ordered list and agree it is easier to separate the two long names. Looking through the code again.

This makes total sense and thanks for looking again!

mads-bertelsen (approved):
Looks great, found no issues!

PR #2571, opened 2026-07-15
[NOT READY FOR MERGE] Union refactoring of scattering while loop

Lomholy:
### Free-form text area

This PR aims to improve the readability of the Union_master component. The PR is limited in scope to only the scattering while loop inside Union_master.comp.

To perform this refactoring I propose two things:
1. We separate all non-primary logic paths, out into separate functions (Currently placed in Union_master for my own simplicity, but will be moved to union-lib.c).

2. When determining a non-primary logic paths, we use wrapper functions, that clearly indicate the intent of the branching in logic path.
An example of this is using:
`
        if (volume_is_only_absorber(Volumes[current_volume])) {
`
Instead of
`
        if (Volumes[current_volume]->p_physics->number_of_processes == 0) { // If there are no processes, the volume could be vacuum or an absorber
`
`
          if (Volumes[current_volume]->p_physics->is_vacuum == 0) {
`

3. (Maybe) We attempt to adhere to a standardized naming scheme for the functions implemented, such that all inhomogenous paths are called  "inhomogenous_insert_what_function_does();".

--------------
* [ ] Please add a checkmark here if you used AI-tools during the work for this contribution
* Further, please describe how / where and for what the tools were used:

--------------
MacOS Tahoe 26.5.2

--------------
  * [ ] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the component (please attach as screenshot in comments!)
  * [ ] I have ensured that basic use of the component is OK (e.g. an instrument using it compiles?)
* ### My contribution includes patches to an **existing** instrument file
  * [ ] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the instrument (please attach as screenshot in comments!)
  * [ ] I have used the `mctest` utility to **test** the instrument (please attach `mcviewtest` report as screenshot in comments)
  * [ ] I am have added reasoning and documentation for the change below

--------------


mcstas-comps/union/Union_master.comp
```diff
@@ -83,6 +83,278 @@ SHARE
...
+
+    *abs_weight_factor = exp (-Volume->p_physics->my_a * 2200 * *time_to_boundary);
+    *abs_weight_factor_set = 1;
+
+    #ifdef Union_trace_verbal_setting
+    printf ("name of material: %s \n", Volumes->name);
+    printf ("length to boundery  = %f\n", length_to_boundary);
+    printf ("absorption cross section  = %f\n", Volumes->p_physics->my_a);
+    printf ("chance to get through this length of absorber: %f %%\n", 100 * exp (-Volumes->p_physics->my_a * length_to_boundary));
+    #endif
```
mads-bertelsen, line 119: The verbal setting will need to change if we support it within union-lib.c, as that setting is currently used by editing Union_master.comp, which is read after union-lib and thus none of the trace_verbal sections of the union-lib.c would trigger. It is better to set this in the instrument, which is possible with a define in the right place, just needs to be documented.

mcstas-comps/union/Union_master.comp
```diff
@@ -83,6 +83,278 @@ SHARE
...
+  }
+
+  void
+  choose_scattering_point_and_ray_direction (double* forced_length_to_scattering, double* safety_distance, double* safety_distance2, double* length_to_boundary,
```
mads-bertelsen: As far as I remember this code only runs when a process that requires a forced position due to focusing is included, but the name of the function sounds like the main way of choosing scattering point. Should make that clear in the name.

mads-bertelsen (commented):
This is an ongoing review, but want to get some information across!

The trace loop is certainly easier to read with this. Naming the functions in this complex system is difficult, will think more about how exactly that would be the most intuitive. It's mainly confusing when reading the functions before reading the trace section, which of course is not often done and can be alleviated with a header.

The function calls use pointers for everything, that reduce the overhead on function calls as nothing needs to be copied. It does make the syntax in the functions a bit harder to understand due to all the dereference operations. When a parameter is not updated in a function, one can make a local variable for the dereferenced value and use that in the equations, it should be optimized out by the compiler while being slightly more readable.

Even with the fast function calls, I do want to check the performance impact, although I don't expect it to be significant.

Lomholy:
> The trace loop is certainly easier to read with this. Naming the functions in this complex system is difficult, will think more about how exactly that would be the most intuitive. It's mainly confusing when reading the functions before reading the trace section, which of course is not often done and can be alleviated with a header.

This is true. We might also think about grouping them together, such that all functions "belonging" to a type of logic, sit close to each other in union-lib. Looking forward to hearing your ideas regarding naming!

> The function calls use pointers for everything, that reduce the overhead on function calls as nothing needs to be copied. It does make the syntax in the functions a bit harder to understand due to all the dereference operations. When a parameter is not updated in a function, one can make a local variable for the dereferenced value and use that in the equations, it should be optimized out by the compiler while being slightly more readable.

I was considering if we should make som helper structs, such that the function calls inside the union_master wouldn't have to be so lengthy. By doing this I also believe these dereferencing issues could be adressed to some degree.

> Even with the fast function calls, I do want to check the performance impact, although I don't expect it to be significant.

Sounds like a good plan. You never know what operations you are by accident performing!

mcstas-comps/union/Union_master.comp
Lomholy: Is the idea for a common user to be able to use this verbal settings?

If so then it definitely should be in the instrument. But, otherwise wouldn't it be fine to make a ``#define Union_trace_verbal_setting`` inside the union-lib.c, if the intent is that it is a debugging tool for the developers?

mcstas-comps/union/Union_master.comp
Lomholy: Will update the name of it! It could be something like ``void focusing_process_choose_scattering_point_and_ray_dir()`` I suppose?
Should I wait until you've reviewed a bit more, or how do we do this efficiently?

mcstas-comps/union/Union_master.comp
mads-bertelsen: The verbal setting is intended to be for super-users, which is a compromise with performance as that avoids a large number of if statements. With a little documentation it will be fine, it's best to enable / disable in the instrument so the handle to do that by removing a comment within Union_master should be removed. Just a little thing that needs doing in connection with this change.

mcstas-comps/union/Union_master.comp
mads-bertelsen: You can update the branch as you want, it's on me that it will take a little longer if I don't finish in one go!

The only processes that use the forced position are the ones that focus during calculation of cross section, so perhaps `focus_in_crosssection_set_forced_point_and_dir()`. It is hard to make a good scheme with the amount of paths through that part of the code.

mcstas-comps/union/Union_master.comp
```diff
@@ -83,6 +83,278 @@ SHARE
...
+  }
+
+  int
+  process_needs_inhomogenous_sampling (struct physics_struct* current_p_physics, struct scattering_process_struct* process) {
+    if (current_p_physics->sampling_points != 0 && process->needs_cross_section_focus) {
+      if (process->sampling_points != -1)
+        return 1;
+    }
+    return 0;
+  }
```
mads-bertelsen: Why does this depend on the `needs_cross_section_focus` part of the struct?

Do we even support processes that need focusing in the cross section calculation, and thus get a forced scattering position to use if the material choose that process, simultaneously being inhomogenous?

mcstas-comps/union/Union_master.comp
```diff
@@ -83,6 +83,278 @@ SHARE
...
+    *my_sum_plus_abs = Volume->p_physics->my_a * (2200 / *v_length);
+    *length_to_boundary = *time_to_boundary * *v_length;
+
+    *abs_weight_factor = exp (-Volume->p_physics->my_a * 2200 * *time_to_boundary);
+    *abs_weight_factor_set = 1;
+
+    #ifdef Union_trace_verbal_setting
+    printf ("name of material: %s \n", Volumes->name);
+    printf ("length to boundery  = %f\n", length_to_boundary);
+    printf ("absorption cross section  = %f\n", Volumes->p_physics->my_a);
```
mads-bertelsen: There is no access to the Volumes array within this function, I believe this would just fail

mcstas-comps/union/Union_master.comp
```diff
@@ -83,6 +83,278 @@ SHARE
...
+  mu_and_intersect_dist_safeguard (double mu_sum, double length_to_boundary, double safety_distance2, int* scattering_event) {
+    if (mu_sum < 1E-18) {
+      scattering_event = 0;
+      return 0;
+    }
+    if (length_to_boundary < safety_distance2) {
+      scattering_event = 0;
+      return 0;
+    }
+    return 1;
```
mads-bertelsen, line 183: Since scattering_event is a pointer, setting it to 0 doesn't really make sense, it needs to be dereferenced first

mcstas-comps/union/Union_master.comp
Lomholy: It does so, because if a material contains both an inhomogenous process, and a focusing process, then the focusing process will need to be sampled as often as the inhomogenous sample.

 > Do we even support processes that need focusing in the cross section calculation, and thus get a forced scattering position to use if the material choose that process, simultaneously being inhomogenous?

By this, I take it you mean a material that uses both an inhomogenous process and a focusing process. If this is the case then I am quite sure that yes, we do support it (It was at least my intention to do so!)

But still there is an error here. The if statement should be separated, such that we first check if the material has sampling points, then if the process is either inhomogenous, or needs cross section focus, we should return 1.

mcstas-comps/union/Union_master.comp
Lomholy: Now dereferenced. Thank you!

mcstas-comps/union/Union_master.comp
Lomholy: Changed access to Volume struct instead of Volumes array

mcstas-comps/union/Union_master.comp
Lomholy: I updated the name to ` focus_in_cross_section_set_forced_point_and_dir ` It may be an excessively long name, but alas so it must be.

mcstas-comps/union/Union_master.comp
Lomholy: I moved this function all the way up (line 97), but I have implemented the logic in the comment above.

mcstas-comps/union/Union_master.comp
Lomholy: Should we make a Test_Union_master_verbose.instr example instrument that does this? 

I just tried running a #define Union_trace_verbal_setting from the declare scope of an instrument, and this  seems to be the correct position for it.

willend: @Lomholy I think it is maybe a bit much to add a whole instr for that? All that is required for verbosity to be compiled in is 

```c
DEPENDENCY “ -DUnion_trace_verbal_setting “
```

in the instr file… - But an option would be to include that commented, i.e. something like

```c
// Uncomment the below line to enable verbosity in Union master - e.g. for debugging purposses
// DEPENDENCY “ -DUnion_trace_verbal_setting “
```

Lomholy: @willend good point. Adding an entire test instrument for one specific piece of documentation is misplaced.

Since it is such a small thing to include in an instrument, I propose that we add a few lines about it in the Union_master header:
```
    * Algorithm:
    * Described elsewhere
    *
    * Debugging:
    * Union_master includes three parameters for debugging through its component interface. 
    * These debugs should be enough for most users. However, if it is necessary to debug the trace loop
    * of Union_master.comp this can be done by adding the following dependency to your instrument file:
    * DEPENDENCY “ -DUnion_trace_verbal_setting “
    *
```

What do you guys think?

willend:
@Lomholy the macOS test will very likely run through OK if you update the branch against current main.

(Seems to be a specific issue with mpi that I will resolve separately later.)

willend:
@Lomholy and @mads-bertelsen I've just updated against main. Is the status still [NOT READY FOR MERGE]?

Lomholy:
> @Lomholy and @mads-bertelsen I've just updated against main. Is the status still [NOT READY FOR MERGE]?

@willend I am seeing @mads-bertelsen next week, where I think we will have a chat on it! Will update the PR after this.

willend:
@Lomholy just fine - had a chat with @mads-bertelsen about this also.

Also, we agreed to use the 'Draft' button for anything that is still 'work in progress' - so have flipped the switch on this PR in that direction.

BTW. McStas 3.8.0 is almost there - this union update will just wait a little / come post that release. :-)

PR #2591, opened 2026-08-07
[Ref] move static comparison out of hot TRACE path

### LINKED ISSUE #2590: Static non-trivial `WHEN` statements should not be evaulated in `TRACE`
The Union test instrument `Test_inhomogenous_process.instr` has a string valued Instrument parameter `sample` which is used to control the possible interactions of a neutron ray in the simulation runtime.
The `sample` Instrument parameter is static during the lifetime of the simulation, so the result of the following check is static as well
https://github.com/mccode-dev/McCode/blob/7f0994cc95a3f213f2297f3da39fc9aaf7311bde/mcstas-comps/examples/Tests_union/Test_inhomogenous_process/Test_inhomogenous_process.instr#L194

Since the `WHEN` expression is evaluated in `TRACE`, this same expression is run as many times as there are rays, e.g., `mcncount`.
A compiler may be clever enough to recognize that the result of this expression is constant, but we can help it by moving such statements to the `INITIALIZE` block of the instrument file.

g5t:
## Do not perform string comparison in `WHEN` statement
The comparison of a static string against a string valued instrument parameter is constant over the lifetime of a simulation. 
Comparing it once per simulated ray is unnecessary and can be 

--------------
* [ ] Please add a checkmark here if you used AI-tools during the work for this contribution

--------------
Linux

--------------
* ### My contribution includes patches to an **existing** instrument file
  * [x] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the instrument (please attach as screenshot in comments!)
  * [x] I have used the `mctest` utility to **test** the instrument (please attach `mcviewtest` report as screenshot in comments)
  * [x] I have used the `mcrun --c-lint` "linter" and followed advice to remove most / all warnings that are raised
  * [x] I have indicated the issue number here:

--------------

Fixes #2590

### DIFF
--- a/mcstas-comps/examples/Tests_union/Test_inhomogenous_process/Test_inhomogenous_process.instr
+++ b/mcstas-comps/examples/Tests_union/Test_inhomogenous_process/Test_inhomogenous_process.instr
@@ -138,7 +138,7 @@ INITIALIZE
     activate_gauss = 1;
     width = 0.01;
   } else {
-    fprintf (stderr, "ERROR(%s): Unsupported parameter value sample=\"%s\".\n", NAME_CURRENT_COMP, sample);
+    fprintf (stderr, "Unsupported instrument parameter value sample=\"%s\".\n", sample);
     exit (-1);
   }
 
--- a/mcstas-comps/examples/Tests_union/Test_inhomogenous_process/Test_inhomogenous_process.instr
+++ b/mcstas-comps/examples/Tests_union/Test_inhomogenous_process/Test_inhomogenous_process.instr
@@ -82,6 +82,7 @@ DECLARE
 %{
   double E_i;
   double eps;
+  int is_thin;
   int activate_inc;
   int activate_inho;
   int activate_inho_linear;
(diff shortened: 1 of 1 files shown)


g5t:
I'm not uploading screenshots because

1. `mcdoc` output is unchanged
2. `mctest` passed but `mcviewtest` fails on line 254, `iterate_obj_to_populate_rows` :shrug:

willend:
@g5t it’s admittedly not so well documented but but we have `#define NAME_INSTRUMENT (instrument->_name)` that can be used in the instrument scope in place of `NAME_CURRENT_COMP`

willend:
> I'm not uploading screenshots because
> 
> mcdoc output is unchanged
> mctest passed but mcviewtest fails on line 254, iterate_obj_to_populate_rows 🤷

@g5t this is fair. Point me to the test artifects and I might have a look early next week.

willend:
@g5t I’ll just merge. You probably attempted to run mcviewtest locally? It is a little picky, it works best if the output dirs from mctest are added to an empty folder - it tries to use “oldest folder available” as “ref” column, and if by chance that folder is from something other than mctest it will fail like you mention.

PR #2608, opened 2026-08-26
Add instrument examples of SPODI (FRMII) powder diffractometer

kinetik161:
### Free-form text area
Here are two models of the SPODI (FRMII) powder diffractometer:

- a model using a simplified banana detector: `FRMII_SPODI`
- a model representing the real multi-detector geometry: `FRMII_SPODI_MULTI.instr`

I have not added anything else except for instrument files themselves and documentation `README` files. @willend , could you please check whether the models in their current form are sufficient?

--------------
Developed and tested on Windows

--------------

* ### My contribution includes a **new instrument** file
  * [x] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the instrument (please attach as screenshot in comments!)
  * [x] I have ensured that basic use of the instrument is OK (e.g. it compiles?)
  * [x] ... and provided reasonable default parameters in that instrument that produce reasonable output
  * [x] ... and maybe even added a `%Example:` line to describe expected behaviour
  * [x] I have used the `mcrun --c-lint` "linter" and followed advice to remove most / all warnings that are raised
  * [x] My new instrument is added within the `examples` hierarchy in a folder in the style of `examples/ESS/New_stuff/New_stuff.instr`
  * [x] My new instrument has a new, unique filename, not clashing with existing example instruments
--------------


willend:
@kinetik161 yes will have a look as soon as time allows. :-)

willend:
@kinetik161 rebased to include various upstream tool / ci changes

willend:
@kinetik161 I think this looks like good work - thanks for contributing!

Only minor things left to do:
1. Include an `%Example: ` line (at minimum in the simple variant where e.g. `theta_narrow` or `theta_full` output is a reasonable single-monitor target). 
  * An easy way to get this information is to run with `-y` parameter defaults and grep for the detector output, e.g.  `mcrun -c --mpi=auto FRMII_SPODI.instr -y -n1e8 | grep theta_full` - I got this line: 
  `Detector: theta_full_I=31699.3 theta_full_ERR=38.2526 theta_full_N=3.60193e+06 "theta_full.dat"`
  * (double-check with a few simulations of varying statistics, to check the stability of the value - do an average.)
2. I would probably add a `SPLIT` on the mono also - the default of x10 should be fine.
3. Please also run the linter. :-) (Time-consuming, may not show anything in your case and a little annoying - but may provide useful information/things to fix.)

kinetik161:
Thanks for the review and recommendations.

1. I have added an `%Example` for the simplified version using values averaged from runs with `NCOUNTS` of 1e8. Should one specify that explicitly in the `%Example` with `-n1e8`? With lower counts, the value varies a bit noticeably due to the long simulation path.

2. Thank you, with split trajectories on the monochromator, the values look much more reasonable at the standard 1e6. I have added that for both models.

3. I can confirm that the linter (`mcrun --c-lint`) passed without any warnings, and that it is extremely slow :)

There is a warning at the generation stage, though I suppose it is more general rather than related specifically to the instrument:
```text
WARNING:
 The parameter format of sample_holder is initialized
 using a static {,,,} vector.
  -> Such static vectors support literal numbers ONLY.
  -> Any vector use of variables or defines must happen via a
     DECLARE/INITIALIZE pointer.
 ```

willend:
These are general warnings indeed, no issue. :-) Hereby accepted and thanks for contributing!

PR #2620, opened 2026-09-01
Fix MT random seed and SPLIT KISS random state restore

mads-bertelsen-agentic:
### Free-form text area

While looking around in the codebase I think I stumbled over a couple of problems with the random number generation that I here propose a fix for. This was discovered while using AI to analyze the current systems. Let me know if I misunderstood something and any of these fixes are actually necessary!

1: Mismatch in defines with MT causing seed to be ignored

mccode/src/cogen.c.in:2422-2424
```
#define _RNG_ALG_MT   1
#define _RNG_ALG_KISS 2
```

MT initialization is performed only by:
mccode-r.c:4259-4270
```
void mt_srandom(uint32_t seed) {
    mt[0] = seed;
    ...
}
```

mccode_main.c
```
#if RNG_ALG == 2
  mt_srandom(mcseed);
#endif
```
The above RNG_ALG sets the MT seed only when KISS is used instead of MT. Simple fix of RNG_ALG == 1 instead.

2: When the KISS random number generator is used (default), SPLIT needs to copy and restore the random state of each ray to continue the random number generation instead of redoing the same RNG. The KISS random number generate has 7 values on the particle struct, but only the first is copied / restored, so the quality of random numbers in SPLIT is reduced, easy fix to loop over the included values. 

Questions in draft stage:
- Does this change in cogent.c.in constitute a change in the code generator that need a design decision entry?
- Test of MT, I did the run below, is that the correct approach?

```
MCSTAS_CFLAGS_OVERRIDE="-DRNG_ALG=1" \
mctest --instr=Unittest_SPLIT --testdir=run_split_mt --ncount 1E7 --limit 35 --skipnontest --mpi=3 
```

--------------

--------------

--------------
  * [ ] I have added reasoning and documentation for the change through an ADR record in our [GRAMMAR](https://github.com/mccode-dev/McCode/tree/main/docs/GRAMMAR) section
  * [X] I am have added reasoning and documentation for the change below
  * [ ] I have indicated the issue number here:

--------------


mads-bertelsen:
Checking the failed tests I see several of them are templateSANS_mantid which does use SPLIT (without a number). I am looking into if this instrument actually produce significantly different results before and after this patch.

mads-bertelsen:
It seems that the templateSANS_mantid test failing is just a fluke, it converges to the target at higher ncounts and using random seed, but the tests have a fixed seed, so it will always fail here. The random number generation in SPLIT is changed with the PR, so it shouldn't be identical to the old results.

willend:
I will 
* rebase
* see if I can modify the test value slightly so that the default seed / value converges
* run a full test-suite to check if there side-effects anywhere else in the testsuite

willend:
@mads-bertelsen I can't seem to start the mcstas-conda-testsuite workflow from the @mads-bertelsen-agentic branch, can you trigger that to run manually just once?

willend:
@mads-bertelsen I am running a 4-core run 1e7 on my local mac for now.

willend:
Local test OK except a curl retrieval of an mcpl.gz for ESS_butterfly_MCPL_test.

Screenshot 2026-09-07 at 16 27 15

(And this a fix for that is under way at https://github.com/mccode-dev/McCode/pull/2630)

Deemed OK for hockey - merging.

Issue #2621, opened 2026-09-02
Instance `COPY` empty-`EXTEND` bug -> feature


g5t:
In the grammar rules for `COMPONENT duplicate = COPY(source)` 
https://github.com/mccode-dev/McCode/blob/03e2db038115278958ff235ad3803209b1ad58e8/mccode/src/instrument.y#L1280-L1314
the resulting instance's `extend` is set to be the same as the source instance.

When translating and compiling the example instrument `RTP_Laue.instr` warnings are issued to alert the user that a copied component instance's `EXTEND` block is overwritten:
```
-----------------------------------------------------------
WARNING: Existing (COPY) EXTEND block in COMPONENT image_plate_eff:
  EXTEND %{
    /* take into account ZnS scintillator efficiency */
    double eff=0.5;
    double v=sqrt(vx*vx+vy*vy+vz*vz);
    p *= 1.0-exp(3960.0/v*log(1.0-eff/100.0));
  %}

is overwritten by:
  EXTEND %{
  %}
-----------------------------------------------------------

-----------------------------------------------------------
WARNING: Existing (COPY) EXTEND block in COMPONENT image_plate_scattered:
  EXTEND %{
    /* take into account ZnS scintillator efficiency */
    double eff=0.5;
    double v=sqrt(vx*vx+vy*vy+vz*vz);
    p *= 1.0-exp(3960.0/v*log(1.0-eff/100.0));
  %}

is overwritten by:
  EXTEND %{
  %}
-----------------------------------------------------------
```
*despite* the file not specifying _any_ extra `EXTEND` on the copies 
https://github.com/mccode-dev/McCode/blob/03e2db038115278958ff235ad3803209b1ad58e8/mcstas-comps/examples/TRIGA/RTP_Laue/RTP_Laue.instr#L110-L126

This seems to have been recognized previously, as some example instruments work around the problem by defining identical `EXTEND` blocks on copied component instances, e.g.,
https://github.com/mccode-dev/McCode/blob/03e2db038115278958ff235ad3803209b1ad58e8/mcstas-comps/examples/ILL/ILL_H22_D1B/ILL_H22_D1B.instr#L225-L254

From the warning messages, the problem seems to arise from the line number check in
https://github.com/mccode-dev/McCode/blob/03e2db038115278958ff235ad3803209b1ad58e8/mccode/src/instrument.y#L1419-L1443
but the actual 'empty code block' initializer sets that to -1
[...]

PR #2627, opened 2026-09-05
Cogen: Simplify COPY(instance) to not transfer EXTEND et al.

willend:
### Free-form text area

The work in this PR starts from the discussions in https://github.com/mccode-dev/McCode/issues/2621 and 

* re-implements `COPY(instance)` without `EXTEND`.
* For symmetry,  transfer of all of `WHEN`, `GROUP`, `JUMP` and `metadata` is also dropped.
* The warning that "`EXTEND` would be overwritten" has been suppressed.
* An ADR record is included in the PR.
* On top of this minor instrument-edits and improvement of changed instr/comp code in basictest workflows

--------------

AI was used to compile the attached spreadsheet that shows very limited effect (non-tested monitors in `ILL_H5` and `ILL_H5_new`) the changes in the PR are estimated to have. (Feedback from the AI is visible on #2621 and is copied in below.)

[copy_survey 1.csv](https://github.com/user-attachments/files/31862540/copy_survey.1.csv)

> It turns out the actual risk pool is very small.
> 
> I re-walked every .instr file in component-declaration order, resolved each COPY(source) / COPY(PREVIOUS) / COPY(PREVIOUS(n)) back to its source (including chains — a copy of a copy), and compared "effective value today" vs. "effective value with inheritance removed" for GROUP, JUMP, WHEN, METADATA. EXTEND is reported separately and excluded from the risk count per your note that its silent-copy path is already broken.
> 
> <strong>Result — real behaviour-change risk is essentially isolated to one pair of files:</strong>
> keyword | McStas files at risk | McXtrace files at risk
> -- | -- | --
> GROUP | 0 | 0
> JUMP | 0 | 0
> WHEN | 2 | 0
> METADATA | 0 | 0
> 
> The two flagged files are ILL_H5.instr and ILL_H5_new.instr (near-duplicate instruments), 3 instances total. Concretely, in ILL_H5.instr line 946:
> 
> COMPONENT H511_mirror_outdXY = COPY(H511_mirror_ReflectXY)(options="dx limits=[-2 2], dy limits=[-2 2]")
> 
> H511_mirror_ReflectXY carries WHEN flag==1, and this copy restates no WHEN of its own — so today it silently only fires when flag==1; once inheritance is dropped it will fire unconditionally, double-counting into that monitor. (A sibling copy in the same block, H511_mirror_TransmitXY, is fine because it does write its own WHEN flag==2.) The same pattern repeats once in ILL_H5.instr and twice in ILL_H5_new.instr.
> 
> GROUP and JUMP came back at zero risk because, looking at how they're actually used across the codebase, authors always restate GROUP explicitly on every group member and always give JUMP its own target — nobody appears to rely on inheriting those silently. METADATA is zero simply because no instance anywhere in either tree writes an instance-level METADATA clause at all yet.
> 
> One methodological note: the resolver found zero unresolved COPY sources in either tree (every COPY target was defined earlier in the same file), so this isn't a partial estimate hedged by missing data — it's a complete, source-order resolution across both codebases, modulo the usual caveats around %include-composed instruments and the rare REMOVABLE prefix (23 occurrences repo-wide).

--------------

--------------

  * [x] I have added reasoning and documentation for the change through an ADR record in our [GRAMMAR](https://github.com/mccode-dev/McCode/tree/main/docs/GRAMMAR) section (`doc/GRAMMAR/ADR-records/ADR_20260904_COPY_INSTRUMENT.md`)
  * [x] I am attaching test output in the comments
  * [x] I have indicated the issue number here: #2621 
  * [x] I have added documentation for the fix and possible side effects

--------------


willend:
A testsuite-output for McXtrace can be found here - all is green:
https://tmp.mccode.org/PR2627/mcxtrace-nightly/mcxtrace-nightly_output.html

willend:
McStas testsuite-output is available here: https://tmp.mccode.org/PR2627/mcstas-nightly/mcstas-nightly_output.html

**Issues:**
* False positive [compilation problem](https://github.com/mccode-dev/McCode/issues/2628) across platforms - this can be ignored, presents itself like this in the log on all test instances in the comparison run:
```
ESS_butterfly_MCPL_test                     : COMPILE ERROR using:
mcrun --mpi=1  --verbose -c -n0 ESS_butterfly_MCPL_test > compile_stdout.txt 2>&1 
```
* The remaining issues are all `mcstas-antlr` related: 
  * Overall status `FAILED! One or more tests errored (1 compile errs / 1 runtime errs / 3 values off` (compile issue is the above false postive)
  * `RTP_Laue                                     :   3.41    [val: 1722.75 / 179284.0 = 1 %] <--- BIG DISCREPANCY?? ` (directly relates to `COPY`+`EXTEND` handling - should be solved on `mccode-antlr` main - i.e. ✅ )
  * `SEMSANS_instrument                           :   1.40    [val: 0.0 / 2.87073e-09 = 0 %] <--- BIG DISCREPANCY??` Requires separate investigation - no use of COPY` or `EXTEND`
  * `BTsimple                                     :   2.17    [val: -1 / 75.4758 = -1 %] + !! RUNTIME FAILURE - see run_stdout_1.txt !!  <--- BIG DISCREPANCY??` includes a "post-processing monitor" in the `FINALLY` of the instr, perhaps something prevents this from functioning in `mcstas-antlr`?

willend:
Failure-state on last commit corresponds to https://github.com/mccode-dev/mccode-antlr/issues/334 (https://github.com/mccode-dev/mccode-antlr/issues/333 is also relevant but not directly reflected in commit-triggered test)

willend:
Gave the testsuite another spin here: https://github.com/mccode-dev/McCode/actions/runs/34024322292 (used  latest `mccode-antlr`)

All McStas runs came out green (warning/error in deployment of mccode-antlr, known issue and no concern).
https://tmp.mccode.org/PR2627/mcstas-nightly-060926-1600/mcstas-nightly_output.html

docs/GRAMMAR/ADR-records/ADR_20260904_COPY_INSTRUMENT.md
```diff
@@ -0,0 +1,38 @@
...
+  EXTEND %{
+  %}
+```
+even in the case of an empty / non-existent EXTEND %{%} block on the copy.
+
+## Decision
+
+For sake of simplicity and to ensure instrument-readability @willend, @g5t and @mads-bertelsen have agreed to no longer include an `EXTEND` from `instance` when `COPY(instance)`. Adding an explicit `EXTEND` block (even if duplicated) is much more explicit.
+
+By symmetry, `METADATA`, `GROUP`, `WHEN`, `JUMP`, and `SPLIT` are from now on also not copied - only `instamnce` and its `instance->actual` are copied.
```
g5t:
```suggestion
By symmetry, `METADATA`, `GROUP`, `WHEN`, `JUMP`, and `SPLIT` are from now on also not copied - only `instance` and its `instance->actual` are copied.
```

docs/GRAMMAR/ADR-records/ADR_20260904_COPY_INSTRUMENT.md
```diff
@@ -0,0 +1,38 @@
...
+
+[`mccode-antlr`](https://github.com/mccode-dev/mccode-antlr/) already implements this behaviour for `EXTEND` since [this commit](https://github.com/mccode-dev/mccode-antlr/commit/7786fd31efde941a16e028f6a616f5b8284c0b50)
```
g5t:
```suggestion
> [!NOTE]  
> This behavior is implement in [`mccode-antlr`](https://github.com/mccode-dev/mccode-antlr/)`>=0.28.0`
```

g5t (changes requested):
The code generator changes look right. I released `mccode-antlr==0.28.0` with the same changes, so the ADR can be updated to reflect that, if you'd like.

docs/GRAMMAR/ADR-records/ADR_20260904_COPY_INSTRUMENT.md
```diff
@@ -0,0 +1,39 @@
+# INSTRUMENT keyword COPY(instance) - only copy instance type and paramters
+
+## Status
+
+*Proposed* and prototype *implemented*
```
g5t: 🤔 I guess this should be `*Accepted*`?

g5t (approved):

docs/GRAMMAR/ADR-records/ADR_20260904_COPY_INSTRUMENT.md
willend: Now reads 

## Status

Change *accepted* and prototype *implemented*

mccode/src/instrument.y
```diff
@@ -1284,15 +1284,16 @@ instref: "COPY" '(' compref ')' actuallist /* make a copy of a previous instance
...
-        comp->group  = comp_src->group;
-        comp->jump   = comp_src->jump;
-        comp->when   = comp_src->when;
         /* now catenate src and actual parameters */
         comp->actuals= symtab_create();
         symtab_cat(comp->actuals, $5);
         symtab_cat(comp->actuals, comp_src->actuals);
-        comp->metadata = metadata_list_copy(comp_src->metadata);
+	/* All other proprties are initialized from fresh */
+	comp->extend = codeblock_new();
```
mads-bertelsen: Looks like this line could be indented one more level to the next indent level

mccode/src/instrument.y
```diff
@@ -1304,12 +1305,13 @@ instref: "COPY" '(' compref ')' actuallist /* make a copy of a previous instance
...
         comp->setpar = comp_src->setpar;
         comp->def    = comp_src->def;
-        comp->extend = comp_src->extend;
-        comp->group  = comp_src->group;
-        comp->jump   = comp_src->jump;
-        comp->when   = comp_src->when;
         comp->actuals= comp_src->actuals;
-        comp->metadata = metadata_list_copy(comp_src->metadata);
+	/* All other proprties are initialized from fresh */
+	comp->extend = codeblock_new();
```
mads-bertelsen: Again here it seems the next indent level would be more appropriate

mads-bertelsen (approved):
Looks great, only comments is some inconsistent indentation (as shown on GitHub), does the clang format work on that type of file? That indent does not block merging!

mccode/src/instrument.y
willend: Artefact from my editor it seems - replaced tab by spaces and it realigns.

mccode/src/instrument.y
willend: Artefact from my editor it seems - replaced tab by spaces and it realigns.

PR #2638, opened 2026-09-09
Patch issue #2637 - Vertical_Bender gravity always disabled

### LINKED ISSUE #2637: [BUG]: Vertical_Bender (contrib): dangling if/else silently disables gravity regardless of -g
**Describe the bug**
In `mcstas-comps/contrib/Vertical_Bender.comp` (INITIALIZE), a dangling if/else with no braces means 
`localG = coords_set(0, 0, 0); `
runs unconditionally, zeroing a correctly-computed gravity vector even when mcgravitation is true. The component silently behaves as if gravity is off, regardless of `-g`.

**Document steps to reproduce the bug**
1) Run any instrument using `Vertical_Bender`, once with `-g` and once without.
2) Results are identical either way — gravity has no effect inside the component.

**Expected behavior**
Gravity should affect the neutron's trajectory inside the bender when enabled, as documented in the component's own header.

**Screenshots**
N/A

**OS and McStas/McXtrace version (please complete the following information):**
 - OS: any
 - McCode variant: McStas
 - Version: bug present since 3.2.1 (when the component was added to contrib); still present in current main 

**Add your code / dataset**
N/A — one-line logic bug, reproducible from the component source alone.

**Additional context**
Fix is trivial (add the missing braces). PR incoming.

MilanKlausz:
### Free-form text area

Fixes #2637. Vertical_Bender.comp's `INITIALIZE` had a dangling `if`/`else` with no braces: 
`localG = coords_set(0, 0, 0); `
ran unconditionally after the `else`, silently zeroing a correctly-computed gravity vector even when mcgravitation was true. This meant the component behaved as if gravity were always off, regardless of `-g`. Fix adds the missing braces so both statements are properly scoped to the else branch. One-line, behavior-only-changes-when-gravity-is-actually-enabled fix — verified to have zero effect on gravity-off simulations (bit-identical output before/after) and a real, expected effect once gravity is genuinely active inside the bender.
--------------

Used Claude (Anthropic) throughout: to trace the bug's root cause (identifying the dangling-else pattern and confirming its scope via the component's other mcgravitation checks), to trace its git history back to its introduction in 2017, and to benchmark its effect before/after the fix on a real instrument model. The fix itself is a two-character change (adding {/}).

--------------
macOS (Darwin). No special dependencies.

The identical fix was extensively tested via a separate instrument model (LOKI) using this component under McStas 3.5.42 / 3.7.24 

--------------
  * [ ] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the component (please attach as screenshot in comments!)
  * [x] I have ensured that basic use of the component is OK (e.g. an instrument using it compiles?)
  * [x] I have indicated the issue number here: #2637

--------------

### DIFF
--- a/mcstas-comps/contrib/Vertical_Bender.comp
+++ b/mcstas-comps/contrib/Vertical_Bender.comp
@@ -214,9 +214,10 @@ INITIALIZE%{
 //	      "This component likely does not work with Gx component,\n",
 //	      NAME_CURRENT_COMP);
 
-  } else
+  } else {
     fprintf(stdout,"Vertical_Bender %s: Gravity is off!\n",NAME_CURRENT_COMP);
     localG = coords_set(0, 0, 0);
+  }
 
     // To be able to handle the situation properly where a component of
     // the gravity is along the z-axis we also define entrance (in) and


willend:
@MilanKlausz and @ your_claude - good catch! :-D

PR #2666, opened 2026-09-17
Refactor External contributions

g5t:
## Record external contributions where they land

`ChopperLib.cmake` fetched `mcstas-chopper-lib` for inclusion of external contributed library and components in the McStas distribution. The use of a bespoke CMake module meant that the record of where those files come from was separate from where a user might look for their definitions.

Each contribution now declares itself in a JSON manifest named `<name>.ext`, **placed in the directory its files are populated into**:

| Manifest | Populates |
| --- | --- |
| `mcstas-comps/share/chopper-lib.ext` | `chopper-lib.c`, `chopper-lib.h` |
| `mcstas-comps/contrib/chopper-lib.ext` | the two `.comp` files |
| `…/examples/Tests_optics/<Instr>/<Instr>.ext` | that one instrument |

`cmake/Modules/External.cmake` reads them: one `mccode_install_externals()` call per directory, mirroring the `install( DIRECTORY ... )` rule already there. Files come from individual URLs or a release `tar.gz` (unpacked once, shared across manifests naming the same one), and every `sha256` is mandatory.

`buildscripts/mcext` computes the hashes (`update` a manifest at a new tag) and re-checks them against upstream (`check`, non-zero exit on mismatch for CI use). Format and options: `docs/EXTERNAL-CONTRIBUTIONS.md`.

### Two commits, deliberately split

- **Populate external contributions from `*.ext` manifests** — the mechanism.
- **Raise the CMake floor from 3.17 to 3.19** for `string( JSON )` . Kept
  separate so it can be reverted alone if something downstream is stuck below 3.19;
  reverting only makes the declared minimum wrong, it doesn't break the build on a
  3.19+ toolchain. The `cmake_policy( VERSION 3.17.0 )` pins under `tools/` are
  left alone on purpose.

### Notes

- Install layout is unchanged; manifests are not installed (`PATTERN "*.ext" EXCLUDE`).
- Offline builds: pre-populate `MCCODE_EXTERNALS_CACHE` or point
  `MCCODE_EXTERNALS_LOCAL` at the files, then set `MCCODE_EXTERNALS_OFFLINE=ON` so a
  gap fails the configure instead of quietly downloading.
- Also clears the CMake 4.x `FetchContent_Populate(chopperlib) is deprecated` warning.

## 
--------------

Claude Code used Opus 5 to implement this external contributed fetch refactor.

--------------

Developed on Linux.

--------------

--------------


willend:
@g5t it looks good but current bandwidth is all reserved for https://github.com/mccode-dev/PaNRAID/blob/main/README.md next week... I will try to have a look at this detail late in / after next week...

mads-bertelsen:
Helpful that it's clear where files are pulled in when browsing the repo. Could you clarify how the hashes are updated when the external repo is updated? I see the mcext can do this work, but is that executed locally and then a pull request started by the developer of the external repo?

g5t:
Following up on the question of how the file hashes get updated: I've pushed a commit that answers it, so that no one has to update them by hand.

- **`buildscripts/mcext` moved out** to [mccode-dev/external-registrar](https://github.com/mccode-dev/external-registrar). The tool that computes and verifies hashes now lives there, together with a GitHub Action that runs it. McCode and upstream repositories use one implementation instead of two copies. Nothing in the build used the script, since `External.cmake` checks hashes itself.
- **New `.github/workflows/external-contributions.yml`**, which runs daily:
  - **poll** finds the latest release of each repository named in a manifest's `git`, and opens a PR from `external/OWNER-REPO` that moves its manifests to that release. The hashes are computed from the files as published. Upstream repositories need no secrets and no setup; they just publish releases. Closing such a PR without merging declines that release.
  - **check** re-verifies every recorded hash against upstream and fails on any mismatch.
- **The poll job needs a `mccode-dev` GitHub App**, whose key stays in this repo. The job is skipped until `REGISTRAR_APP_ID` is set, so this PR can merge before the app exists. The registrar README has the setup steps -- [some of which I've already done](https://github.com/organizations/mccode-dev/settings/apps/mccode-registrar).
- **The branch is rebased** onto the merge of `main`. The CMake 3.19 floor came in through #2668, so the PR no longer carries that change.
- **`chopper-lib` has been updated to include a new `Polygon_ESS_butterfly`.comp` but stays pinned at v4.1.0** in this PR. Once the app is configured, the poll job will propose v4.2.1 by itself; a dry run already finds it.

g5t:
I think this is actually ready now, but obviously want everyone to agree merging is a good idea first.
I've setup the necessary GitHub App and put its `client id` and `secret` in the repository settings, so once this _is_ merged, the workflow can (and should) be tested ASAP.

To that end, these are Opus' instructions for whomever pushes the merge button:

## First run of "External contributions" (dry run)

Do this right after #2666 is merged.

1. **McCode → Actions → External contributions → Run workflow**. Leave the branch as `main` and **only** blank, tick **dry-run**, then click **Run workflow**.
   Or from the command line: `gh workflow run external-contributions.yml -R mccode-dev/McCode -f dry-run=true`
2. Once the run finishes, open it and check the summary:
   - **check** is green and reports `5 manifest(s) verified against upstream`.
   - **poll** is green and its table shows `mcdotstar/mcstas-chopper-lib | v4.1.0 | v4.2.1 | dry-run`.
3. If **poll** shows as *skipped*, the `REGISTRAR_CLIENT_ID` variable isn't set (Settings → Secrets and variables → Actions → Variables).
4. If anything is red, post the run's link in #2666 and don't run it without dry-run.

If all is well, run it again without dry-run. It should open a PR "Register mcdotstar/mcstas-chopper-lib v4.2.1".

mads-bertelsen:
Makes sense that an action helps with the bookkeeping, so we effectively get new versions of the external code in our usual github tests, which provide a good image of what a release would look like. Make sure that action can be executed manually as well.

I also want to make sure there is a sensible way to make some breaking change in the format of components across these repos. Say we change how parameters are defined, we would have those changes on a branch in the main repo with a PR. Should we then have a similar PR on the external repo and ensure they are merged at the same time? 

Would it be possible to make a build of mccode that pulls from a given branch from the external repo such that a breaking change could be easily tested?

willend:
@mads-bertelsen if an instrument (either local to McCode itself or from the incoming repo) includes a %Example, that instrument will be both a) compiled and b) tested using mctest.

g5t:
:thinking: As I see it, whatever we decide amounts to one of three policies:

1. **Forbid breaking grammar changes.**
2. **Standardize how unfixable components are removed** in the PR that introduces the breaking change.
3. **Allow breaking changes without bringing every component up to date.**

Option 1 is the ideal we should aim for. Still, some future change may be worth a `McCode v4`, and in that case removing components that can't be fixed should be allowed. In practice, option 3 is what we do now. See #2678, where a component in `mcstas-comps/contributed` still uses pre-McCode 3 syntax.

I'd rather adopt option 2 on purpose than keep option 3 by accident.

## The case for dropping unfixable components

A breaking change can require so much rework of a component or library that reproducing its old `%Example` output stops being a sensible goal. How we handle that should depend on who owns the component.

**Core components.** The developer of the breaking PR is responsible for fixing these. That's part of the cost of making the change, and CI should enforce it.

**Contributed components.** Here the McCode developers usually aren't the original authors. When we port a contributed component to new syntax or new semantics, we're guessing at the author's intent. We have to guess what a parameter was for, which edge cases mattered, and whether odd-looking code is a bug or a deliberate trick. A port that compiles and runs but quietly changes the physics is worse than no port, because it still carries the original author's name. If we can't get the author, or someone who knows the component well, to check the port, removing it from the new version is safer.

**External repositories.** A library like `chopper-lib` has a life of its own. If a hypothetical `McCode v4` breaks `chopper-lib`, the cleaner route may be a separate `chopper-lib-4` (or a new major version), created if and when someone needs it. Meanwhile the existing library keeps working with McCode 3.

In both of the last two cases, "removed" should mean *marked as unsupported in version N, with a pointer to the last version that works*, not deleted without a trace. Leaving stale components in the tree has its own cost: they break or get excluded from CI, they confuse users who find them in the docs, and they make it harder to tell whether the rest of the library actually works. #2678 shows what happens otherwise: a component nobody fixed just sits in the tree unnoticed.

## Testing a breaking change across repos

The workflow can already be run by hand: `external-contributions.yml` has a `workflow_dispatch` trigger, with inputs to poll a single `OWNER/REPO` and to do a dry run.

It follows *releases*, not branches. That's deliberate, since every manifest records a SHA256 for each file, and a branch that keeps moving would make those hashes stale on every push. The easiest way to test a breaking change against an external repo is therefore to tag a **pre-release** on its matching branch (e.g. `v5.0.0-rc1`) and point the McCode PR branch at that tag:

```sh
mcext update mcstas-comps/share/chopper-lib.ext -v v5.0.0-rc1
```

That repoints the manifest and recomputes its hashes, so the PR's normal CI builds and tests against the external branch.

Merging then goes in this order:

1. Merge the external PR and tag a proper release.
2. Repoint the McCode PR at that release.
3. Merge the McCode PR.

Neither PR needs to wait on the other beyond that, and if the external maintainers decide not to follow the change, the McCode PR drops the manifest instead (see above).

mads-bertelsen:
I consider it a requirement that we can make breaking changes when necessary, of course it's not something we do often and will try to avoid it. Traditionally we have supported both versions for a period.

Your suggestion of using a release candidate on the external repo makes it possible to test breaking changes, had not thought of that option. It would be a bit easier to just specify a target branch on the external repo for testing, but a release can work.

I don't think a new external repo will be needed for each breaking change since its already pinned releases in the mcstas releases, if someone works with an older McStas or even builds it from source, it would grab the external repo components from the matching release?

g5t:
> Your suggestion of using a release candidate on the external repo makes it possible to test breaking changes, had not thought of that option. It would be a bit easier to just specify a target branch on the external repo for testing, but a release can work.

If, in local testing, you find that a breaking change will _also_ break a remote contribution, you can replace any `.ext` file with the remote file(s) and test/modify directly. I guess this would be the best way to work out how to implement breaking changes before opening a issue or PR in the remote repository.
 
> I don't think a new external repo will be needed for each breaking change since its already pinned releases in the mcstas releases, if someone works with an older McStas or even builds it from source, it would grab the external repo components from the matching release?

:shrug: I'd leave that up to the remote repository maintainer(s) -- they may not want to accept the changes for whatever reason. A related issue is what license(s) a remote-contribution is allowed to use to be accepted here; if that is restricted accordingly then forking the remote repository to make the rejected breaking changes would be an option.

mads-bertelsen:
> If, in local testing, you find that a breaking change will _also_ break a remote contribution, you can replace any `.ext` file with the remote file(s) and test/modify directly. I guess this would be the best way to work out how to implement breaking changes before opening a issue or PR in the remote repository.
> 
Sure, thats another way to work that would require a bit more manual work. Could be a good way to figure out what changes would be needed in the external repo before starting to suggest changes there. 

> 
> 🤷 I'd leave that up to the remote repository maintainer(s) -- they may not want to accept the changes for whatever reason. A related issue is what license(s) a remote-contribution is allowed to use to be accepted here; if that is restricted accordingly then forking the remote repository to make the rejected breaking changes would be an option.

That does make sense.. It does make these external contributions more like dependencies. I think we should set requirements on what can be accepted as an external repo in terms of license so that we can fork it if necessary. For now the repositories are within the McCode organization, that could also be a formal requirement for this type of contribution.

I apologize for the many questions, but would really like to understand how this is intended to work in practice and would like to be sure its done in a sustainable way.

PR #2668, opened 2026-09-17
Raise the CMake floor from 3.17 to 3.19

g5t:
Machinery to add *.ext manifests which contain JSON to be parsed by CMAKE with `string( JSON )` requires a minimum CMake version 3.19.

The cmake_policy( VERSION 3.17.0 ) pins under tools/ are deliberately left alone: raising the floor should not quietly change policy behaviour in the directories that explicitly fixed it. Bumping those is a separate decision.

If this PR is accepted, the second commit of #2666, 1603570d5d27f0899458b07ea37911d9904f87b6, can be dropped (though it should be harmless to keep it in place).

--------------

Opus 5 identified all instances of `CMakeLists.txt` which needed a higher minimum `CMake` version.


willend:
As such not much of a formal review needed. - Will let the tests run through and merge.

PR #2675, opened 2026-09-21
Union manual chapters for both of McStas and McXtrace

willend:
### Free-form text area

Union manual chapters for both of McStas and McXtrace
(+ various visual improvements, fixes for big tables etc.)

--------------

Claude was posed with this challenge:

> Inside of docs/Union/union_manual_v0.91.pdf  is a (TeX-based) PDF with documentation for the McStas union framework. Pretty dated.
> 
> Also I give you the related PhD manuscript from Mads Bertelsen. A few years old.
> 
> Please use these - and the actual Union code from mcstas-comps/union and mcstas-comps/share/union*.c - to write me a chapter on Union for the McStas user manual and component manual. Please also look at relevant parts of https://panosc-vinyl.github.io/McStasScript/index.html (but bear in mind that I want the new chapter to relate to basic McStas syntax - not McStasScript syntax)

Followed up with
> I will now try to build. In the meantime you can look at McXtrace Union:
> 
> McXtrace DOES have (a smaller) Union system, find within mcxtrace-comps/union and mcxtrace-comps/share/union*.c
> 

+ once the right structure was in place followed by a series of commits relating to visual appearance of logos, front-pages, tables, broken references etc.

**For reviewers:**
1. Please don't spend any time on the actual TeX / CMake / etc. changes - simply review relevant parts of the generated output. :-)
2. Both of the manual and component manual contain Union material, difference should only be that the comp manual also includes snippets from the comps
3. The manuals (in `.pdf` and `.html` (inside a `.tgz` archive))can be found within `docpkg/manuals/mc*`

--------------

--------------

--------------


willend:
Oh and @mads-bertelsen I forgot to properly credit the two figures that are now placed in the manuals, taken from doc/Union/union_manual_v0.91.pdf - but perhaps you would like something else all together?

mads-bertelsen:
Great job, that has been missing for a decade. There is no need to directly credit me with any of it, thats not really done in other parts of the manual. KU is credited in the component files made during that employment, and ESS for files made in that period, that should be sufficient.

I see no wrong information in the new manual, even though some of the old documentation used as a template  is very old.

There are a few important pieces of information thats missing which were not covered either in the McStasScript webpage or old manual, so it makes sense. These would be nice to add, but not crucial.

- Processes are affected by rotation relative to ABSOLUTE, thats why it's recommended to place them there. The idea was that it would be easy to create a twinned single_crystal by rotating one. I have come to realize this is bad interface, and want to move this functionality to some inputs like rot_x, roy_y, rot_z. 
- More than one surface_process can be attached to a geometry as a stack (but not yet relevant as there is only one process)
- Refraction / Reflection relies on scattering length density being available on the material, so it only happens when both the volume the ray originates from and is trying to move to has that information. The default Vacuum material has refractive index of 1, so that is supported.
- For the absorption loggers it would be nice to mention they can be used to model detectors

mads-bertelsen:
Pushed a small update on the process feature with rotation mentioned in the previous reply because it can be quite confusing. As mentioned I want to redo the feature and can then update the documentation accordingly at that point.

willend:
@mads-bertelsen @ebknudsen if you think this is good enough for merging, can you please add a formal "review" to enable merging?
Screenshot 2026-09-23 at 23 35 38

willend:
Merge-restriction now lifted.

ebknudsen:
Reading through the McXtrace/manual now - recording issues as I find them...
- Figure 7.1 references McStas. Likely it is part of the image-file.

PR #2676, opened 2026-09-21
mcdisplay-webgl: drop Vite/npm from the runtime path

willend:
### Free-form text area

mcdisplay-webgl: drop Vite/npm from the runtime path, serve dist/ stattically

write_browse() no longer shells out to `npm run dev` (Vite dev server) to view a trace - dist/ is a self-contained, hashed Vite production build, so the stdlib ThreadingHTTPServer serves it directly with no Node/npm/vite runtime dependency and no per-launch startup cost.

This removes the whole per-user ~/.mcstas/<ver>/webgl deployment: no more node_modules symlink/junction, no start-vite.js, no first-launch npminstall requiring internet access. mcdisplay.py now always reads dist/ from wherever it's installed (conda env or dev checkout) and only builds it once, in place, if it's ever missing.

npminstall(.bat).in simplified to match: build directly next to the source files instead of copying everything into a per-user folder first.

CMakeLists.txt: drop the empty node_modules install and start-vite.js (unused now); add model/, vite.config.mts, tsconfig*.json so the fallback `npm run build` matches a real dev build; drop nodejs/npm/rsync as a package runtime dependency, since dist/ ships pre-built.

vite.config.mts: remove the vite-plugin-static-copy step - it pointed at placeholder paths ("path/to/particles.json") that never existed as real files, and was always dead since vite.config.mts was never previously shipped to end users. instrument.json/particles.json are per-run outputs write_browse() already writes directly into the served directory.

package.json: move build-only packages (vite, @vitejs/plugin-react, @types/*) into devDependencies; drop the now-unused vite-plugin-static-copy dependency.

Verified end-to-end on macOS: cold `npm install && npm run build` from exactly what CMakeLists.txt now ships, served via the new stdlib server, and a real instrument trace through mcdisplay-webgl.

--------------

Claude galore, first prompt:
_I have a dilemma with my mcdisplay-webgl solution:

For sake of platform-independence, I use the script tools/Python/mcdisplay/webgl/npminstall.in (and its .bat counterpart for Windows) to deploy a set of npm modules into the user's (e.g.) .mcstas/3.99.99_mccode-dev/webgl folder.

The deployment is done with first launch of the mcdisplay-webgl tool and a) requires internet access and b) is time-consuming... (Actually more time-consuming than the build of the whole McStas package)

Can you look through the code / requirements and help me figure out if something can be optimised? 

(One thing I thought about already is to deploy centralised in subfolder tools/Python/mcdisplay/webgl/ under some constraints, e.g. when packaing with conda, then symlinking the 'system files' level set of files in the location of the 'user files'... )_

--------------

--------------

--------------


willend:
Have put in DRAFT mode since this really requires proper, off-line testing on every platform!

PR #2677, opened 2026-09-21
Fix Conics display and particle trace parsing

mads-bertelsen-agentic:
### Free-form text area

mcdisplay of conics did not work. Investigated the problem using AI and found two issues, one in the components mcdisplay section and one in the python tool that reads particle traces.

Correct the vector-mode hyperboloid construction in Conics_EH and Conics_PH by passing the full focal separation to addHyperboloid. The previous equal focal arguments generated degenerate geometry and could produce non-finite display coordinates.

Extend FlowChartParticleTraceParser to recognize LEAVE immediately after SCATTER. Conic components can emit an internal SCATTER trace record through conic.h, followed directly by the instrument-level LEAVE record. The parser previously tested only for STATE or COMP after SCATTER and routed this valid sequence to t_error4. Add a dedicated leave decision and route the sequence through the existing leave-state handler while preserving the old error path for other unexpected records.

It was found that the Conics_He and Conics_PP were not included in the test instrument and had similar issues, these were corrected and the components added as cases to the test instrument. The test instrument was updated as some focal distances were hardcoded even when the instrument parameters were already available.

Verified with 300 captured particle-trace blocks, mcdisplay-webgl in default and OPTIC=2 modes of the Test_Conics_pairs instrument. Mcdisplay screenshots below

Conics_PH
Conics_PH

Conics_EH
Conics_EH

Conics_HE
Conics_HE

Conics_PP
Conics_PP

--------------

--------------
OS X 14.6.1 (23G93)

--------------
  * [x] I have used the `mcdoc` utility and **rendered** a reasonable documentation page for the component (please attach as screenshot in comments!)
  * [X] I have ensured that basic use of the component is OK (e.g. an instrument using it compiles?)
  * [X] I have used the `mctest` utility to **test** one or more instruments making use of the component (please attach `mcviewtest` report as screenshot in comments)
  * [x] I have used the `mccode-clangformat` tool to apply the standard McCode component indentation scheme
  * [X] I have used the `mcrun --c-lint` "linter" and followed advice to remove most / all warnings that are raised

--------------


mads-bertelsen-agentic:
The correction of the geometry in the case where an array of radii is given actually results in loosing half the intensity in the tests which was concerning. The reason turned out the be that both conic pairs were focused on the sample, and thus in the situation with few mirror shells, there was more solid angle covered that reflect towards the sample. Wolter optics should only focus rays hitting the first conic to split the change of direction into two shallower reflections. The Test instrument now includes monitors for each number of reflections, and on the previous version all the intensity was one reflection, while on the new its in two reflections as expected. Furthermore when adding a sufficient amount of nested mirrors, they start shadowing in the earlier wrong geometry, while they continue to scale with the new, this is shown below. 

wrong_vs_right
nshell scan of test instrument with fi=5. Left old given array, right new given array.

We could make separate components that are not pairs but for example just one ellipse that focus point to point, as it seems to be an effective strategy at lower numbers of mirrors.

